Issue #27, opened 2022-01-21
Add template for issues (feature request)


RafaelPalomar:

RafaelPalomar:
This is closed in ALive-research/Slicer-Liver#28

PR #28, opened 2022-01-21
ENH: Add templates for issues

RafaelPalomar:
This closes ALive-research/Slicer-Liver#26 and ALive-research/Slicer-Liver#27

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,31 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Version [e.g. 22]
+
+**Additional context**
+Add any other context about the problem here.
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: enhancement
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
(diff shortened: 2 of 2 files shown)


RafaelPalomar:
What do you guys think about this issue templates?

gatangen (commented):
Suggest to add a label for Slicer version in the bug report. 
Maybe add a separate template for "Improvement  Request" ?

jpdefrutosSINTEF (approved):
Looks good! Together with Geir Arne's suggestion, I'd also add a label for the Liver extension version in use. Perhaps suggest the user to provide a screen recording? or data (if used, and if possible) to reproduce the error? (in addition to the steps needed to trigger the error).

olevs:
This is great. Program versions is nice to have. Perhaps also suggest that the user includes call stack if available? Like in the help text under "Additional context".

olevs (approved):

RafaelPalomar:
> Suggest to add a label for Slicer version in the bug report. Maybe add a separate template for "Improvement Request" ?

@gatangen, isn't that already considered in "feature request", or "improvement request" could be something else?

RafaelPalomar:
> Looks good! Together with Geir Arne's suggestion, I'd also add a label for the Liver extension version in use. Perhaps suggest the user to provide a screen recording? or data (if used, and if possible) to reproduce the error? (in addition to the steps needed to trigger the error).

This we will have to set it up manually ourselves since we are using milestones instead of labels for the version tracking. But definitely we will assign issues to version. The user can specify which version they found the error in.

gatangen:
@RafaelPalomar , I understrand "Feature Request" as a request for new functionality or new characteristics of the system while "Improvement" is more a suggestion for improvement of existing functionality or general improvement of the system. What do you think?

RafaelPalomar:
> @RafaelPalomar , I understrand "Feature Request" as a request for new functionality or new characteristics of the system while "Improvement" is more a suggestion for improvement of existing functionality or general improvement of the system. What do you think?

I see. It makes sense. I have added a new template "Improvement request".

jpdefrutosSINTEF:
After using the bugs template, perhaps I would remove the label "name" and move "title" to the top. I feel these two refer to somewhat the same thing. Something like this:

---
title:
about: 
labels: bug
assignees: ''

---

RafaelPalomar:
> After using the bugs template, perhaps I would remove the label "name" and move "title" to the top. I feel these two refer to somewhat the same thing. Something like this:
> 
> title: about: labels: bug assignees: ''

These are `metadata`. They are not part of the template per se and I think cannot be removed :)

jpdefrutosSINTEF:
> > After using the bugs template, perhaps I would remove the label "name" and move "title" to the top. I feel these two refer to somewhat the same thing. Something like this:
> > title: about: labels: bug assignees: ''
> 
> These are `metadata`. They are not part of the template per se and I think cannot be removed :)

ups!
Then forget what I said ^^

PR #110, opened 2022-03-10
Implementation of Liver vascular segments annotation and calculation

gatangen:


olevs:
I had to delete the github branch as it was corrupted. Uploaded the branch again from my local copy, and made a new pull request: #131

PR #143, opened 2022-06-30
COMP: Fix extension dependencies

RafaelPalomar:


olevs:
I think we have to remove "ExtractCenterline" from self.parent.dependencies in Liver.py as well

Issue #134, opened 2022-05-12
Creation of segments out of different vessels systems


RafaelPalomar:
Now, the centerline extraction for a given set of points works if all those points have been associated to the same model (e.g., 2 points assigned to portal). We would like to expand the use case to include the possibility to add points in different vessels systems.

 As an example, think about  the possibility to add 2 points in portal and 2 points in hepatic for the same point set. At this point, the algorithm for centerline extraction will look to connect the two systems (and will fail because the models should not have any connection).

Idea on how to implement this:
  - Instead of saving list of points, we could save list of tuples (model, point).
  - The algorithm that computes the centerlines will iterate over the number of models added in the list of tuples and will runt the centerline extraction as many times with the relevant points.
  - The id of the centerline (and color) should be the same

olevs:
I've added the functionality for testing out this functionality in the #137 pull request. Now centerline segments are added to the same model instead of being replaced when pressing the "Add Vessel Centerline Segment" button. As long as you use the same markups point list you will be adding segments to the same model. For this first implementation you will need to use the other Slicer widgets to delete markup points or the combined centerline model.

olevs:
Pull request #144 is adding segment id to all segment nodes. This can be used to give segments the same id. No GUI is created for this yet.

PR #154, opened 2022-09-27
Feature/vascular territories

olevs:
User interface changes for creating vascular territories


RafaelPalomar (changes requested):
@olevs, @gatangen, as you can see I have added two new commits:

  - Organization of the GUI in layouts (we should never leave widgets manually placed floating in the GUI). It may affect the resizing of the widget
  - Replacing python `print` by `logging` calls.

Please have a look at those commits and let me know if that's fine. 

Another thing is that there are a lot of commented lines. Some of these might be still useful? But it would be nice to remove  the commented lines that are not going to be used anymore. Python files tend to get lengthy and disorganized, the less lines we have the better.

Apart from this minor things, great work!

RafaelPalomar:
This closes #134

RafaelPalomar (approved):

PR #166, opened 2022-10-17
Fix extension icon

lassoan:
The SlicerLiver extension icon currently did not appear in the Extensions Manager because the URL was not valid.


RafaelPalomar:
Thank you very much, @lassoan. We hadn't notice this.

PR #170, opened 2022-11-04
ENH: Enabled adding more point lists, in order to adding more separat…

olevs:
…e segments to the same liver territory


RafaelPalomar (approved):

RafaelPalomar:
This closes #170.

PR #198, opened 2023-01-30
Unit test framework

olevs:
Added test that covers qSlicerLiverMarkupsModule, vtkSlicerLiverMarkupsLogic, and all LiverMarkups MRML data nodes.


olevs:
@RafaelPalomar: This pull request is now ready to merge, even if there are more tests that needs to be written. This also contain a fix for CircleCI, so it would be good to merge it soon.

LiverResections/qSlicerLiverResectionsModule.cxx
```diff
@@ -126,7 +126,7 @@ QStringList qSlicerLiverResectionsModule::categories() const
 //-----------------------------------------------------------------------------
 QStringList qSlicerLiverResectionsModule::dependencies() const
 {
-  return QStringList() << "LiverMarkups";
+  return QStringList() << "LiverMarkups" << "Markups";
```
RafaelPalomar, line 129: `LiverMarkups` already depends on `Markups` this looks redundant to me. Is there any reason to add it here explicitly @olevs, @gatangen ?

RafaelPalomar (changes requested):

LiverResections/qSlicerLiverResectionsModule.cxx
olevs: The problem was that the python test failed without this addition. We couldn't get the dependencies loaded any other way

RafaelPalomar: I see. Let's keep it then.

PR #200, opened 2023-02-03
ENH: Add distance map downsampling option to save GPU memory

RuoyanMeng:


olevs (approved):

RafaelPalomar:
This PR is superseded by ALive-research/Slicer-Liver#202. The reason is that as is, this branch is conflicting with the `develop`. ALive-research/Slicer-Liver#202 is a conflict-resolved branch that needs to be tested to ensure no errors in the conflict resolution process.

Issue #189, opened 2023-01-19
Default color values


RafaelPalomar:
Right now the default color values (and other settings for that matter) are decoupled between the UI (scripted loadable module) and low lever rendering objects (DM's, Mappers, etc.) We should study the possibility to have these values in a logic in such way that the values can be queried from different components. This will  help keeping the settings consistent.

RuoyanMeng:
Is there any default colormap for the Sliver-Liver extension, and default segments color setting? So I can define the Hepatic/Portal contour color accordingly.

RafaelPalomar:
We could use the `GenericAnatomyColors` color table. Related to this is the use of terminologies (and colors associated) for segmentations and derivatives. Here a recent discussion on this:[https://discourse.slicer.org/t/terminologies-questions/27384]( https://discourse.slicer.org/t/terminologies-questions/27384)

olevs:
In LiverSegments we copy an existing Slicer colormap, and call it SlicerLiverColorMap. See: https://github.com/ALive-research/Slicer-Liver/blob/9502786e8f14f6406296166cd36faa974396160d/LiverSegments/LiverSegments.py#L232
If we did it like this we can add our own colors and labels if necessary.

RuoyanMeng:
The colormap used in LiverSegments is Labels, different from the GenericAnatomyColors we used for representing liver structures. I can hardcode the vessel color now and we can discuss which colormap should be used later.

PR #220, opened 2023-03-17
ENH: Load liver color map from file

### LINKED ISSUE #214: Segment separation leads to segmentations with weird names
You can see how some segments are named: brain or ventricle.

This probably has to do with the color maps

olevs:
Try to fix #214 by loading liver color map from disk. Example:  SlicerLiverColorMap.ctbl

This should also address ALive-research/Slicer-Liver#189


RafaelPalomar:
This is the sequence I would follow:

  - Black 
  - Liver
  - Portal vein
  - Hepatic vein
  - Tumor 1
  - ...
  - Tumor 50
  - Segment 1
  - ...
  - Segment 100

olevs:
Using the fixed numbers in a color table was a bit cumbersome to use. I have kept the color table file, but the color names are now being overwritten by the vascular territory id names from the GUI. This GUI element have now been made editable, so the user can just write the name of the next vascular territory, but maybe we want to change this to rename existing instead?

RafaelPalomar (approved):

RafaelPalomar:
This fixes ALive-research/Slicer-Liver#189 and ALive-research/Slicer-Liver#214

Issue #226, opened 2023-04-17
Scaling factor for resectograms


RafaelPalomar:
Resectograms can be difficult to visualize due to their size. Add a scaling factor (e.g., slider) to scale up the resectogram.

RafaelPalomar:
@RuoyanMeng This is relevant to the discussion we had yesterday towards the user tests

PR #227, opened 2023-04-21
ENH: Arbitrary Resection Volumetry Calculation

RuoyanMeng:
#236, #238 solved


RuoyanMeng:
Hi @RafaelPalomar, I update the UI and functionality: 

1. Now users can calculate ROI volume for any case, with or without resection, using either vascular segments, single resection or multiple resections. 
2. The UI is more intuitive, with basic functions for simple resection volumetry, Use-All-Resections and Vascular-segments as advanced parameters.

Could you try it and give me some feedback?

RuoyanMeng:
The latest commit contains big changes in the code structure and UI. 

- Move the code from live.py and liverResectionLogic to an independent module.
- New design for segmentation and its segments choice, now you can do multi-selection on whatever segments you are interested in.
- Including "Resection" in volume calculation is an advanced feature for volume calculation now. In this version, only "single resection" or "All resection" options are there, will add multi-selection in the future.

RafaelPalomar:
@RuoyanMeng, here there is an example of use of normalized device coordinates for a given actor. (I haved not tried) but this should help you representing your actors in terms of final coordinates in the range 0-1, so you easily scale the resectogram. Let me know when this is ready for review.

https://examples.vtk.org/site/Cxx/Annotation/MultiLineText/

RuoyanMeng:
#226 solved

RafaelPalomar (approved):

PR #232, opened 2023-04-25
ENH: Create new endPointsMarkupsNodes to allow adding disconnected ce…

### LINKED ISSUE #229: Allow adding disconnected centerlines to the same territory-segment combination
Currently each territory-segment combination is automatically generated. If one territory needs more disconnected center lines from the same segmentation/segment another node needs to be created in the "Vessel points" selector. Currently the "Create new Point List" selection just gives the same point list as before. Implement the functionality to create more the one point list for each territory-segment combination.

olevs:
…nterlines to the same territory-segment combination
This fixes #229


olevs:
This closes #229

RafaelPalomar (approved):

PR #230, opened 2023-04-24
ENH: Allow adding whole segmentations to centerline model

olevs:
This fixes #228

PR #233, opened 2023-04-25
ENH: Allow generation of multiple vascular territories maps

### LINKED ISSUE #217: Output of segments separation should be configurable
Right now the output of the segments is placed on the same fixed mrml node. For the surgeons it would be interesting to be able to select the output to have (maybe) different segments distributions that can be compared

### LINKED ISSUE #240: Create vascular territories as segmentations instead label maps
Create vascular territories as segmentations instead of label map volumes

gatangen:
Enable generation of several Vascular Territory label maps, which can be selected from GUI widget.
This fixes #217
Fixes #240


gatangen:
Changes in the widget .ui file will be syncronized with #232 and #230 when merging this PR to develop branch.

olevs (approved):

LiverSegments/Resources/UI/LiverSegments.ui
```diff
@@ -99,111 +59,143 @@
...
+        <item row="0" column="0" colspan="2">
+         <widget class="QLabel" name="label_4">
+          <property name="text">
+           <string>Vascular Territory Segmentation:</string>
           </property>
-         </spacer>
+         </widget>
+        </item>
+        <item row="0" column="2" colspan="3">
+         <widget class="ctkComboBox" name="selectedVascularTerritoryMapId">
```
RafaelPalomar, line 183: This should be a `qMRMLNodeComboBox` (https://apidocs.slicer.org/master/classqMRMLNodeComboBox.html). This would allow the user to create, delete, rename, etc. This will probably imply some minor changes in the logic as you don't have exactly the same methods (e.g., currentIndex is not available).

RafaelPalomar (changes requested):
Consider the use of qMRMLNodeComboBox

LiverSegments/Resources/UI/LiverSegments.ui
RafaelPalomar: One picky detail is that the collapsible button containing all the sub-widgets shows a little shorter than the rest.

RafaelPalomar:
This fixes ALive-research/Slicer-Liver#217
This fixes ALive-research/Slicer-Liver#240

LiverSegments/LiverSegments.py
```diff
@@ -636,9 +676,13 @@ def calculateVascularSegments(self, refVolume, segmentation, centerlineModel, co
 
     labelmapVolumeNode.GetDisplayNode().SetAndObserveColorNodeID(colormap.GetID())
     slicer.util.arrayFromVolumeModified(labelmapVolumeNode)
+    vascularTerritorySegmentationNode.Reset(None)#Existing node will be overwritten
```
olevs, line 679: CircleCI fails, because vascularTerritorySegmentationNode (= int?) got no Reset function

RafaelPalomar:
I ge the following errors in the python console with this PR

```
[VTK] vtkMRMLColorTableNode::GetColor: requested entry -1 is out of table range: 0 - 257, call SetNumberOfColors
[VTK] ERROR: SetColorName, index was out of bounds: -1, current size is 257, table name = SlicerLiverColorMap
```

gatangen:
Thanks Rafael! I will fix the error msg. and update the PR.

RafaelPalomar (changes requested):
- The first name (default) for vessels point is Territory_1_segment_2. Is this expected?
- There is no isolation between parameters for one "vascular territory segmentation" and others. This means that when "vascular territory segmentation 1"  is set, all segmentations from "vascular territory segmentation 2" are available (how to discriminate which one is which?). I think this could be solved by adding a property to all the segmentations generated. This property could be simply the ID of the vascular territory segmentation. Then all the comboboxes can filter by this ID. The result should be that on selecting "vascular territory segmentation 1" only those items generated by it are available in the comboboxes.
- I would not create a vascular territory segmentation by default. I would leave it to none and the user has to create it specifically. In addition I would disable all widgets below until the vascular territory segmentation is created. The same goes for vascular territory ID.
- I cannot create different sets of vascular territories segmentation. The results turned to be the same for the two supposedly different I created.

RafaelPalomar (approved):

PR #239, opened 2023-05-10
Bug/vmtk module check #235

### LINKED ISSUE #235: There are no checks for  ExtractCenterline module presence
**Describe the bug**

Trying to compute vascular segments without the required ExtractCenterline module will not trigger any error. In addition this will render the UI unusable (the 3D view cannot be manipulated and the mouse pointer state is idle). Furthermore, it produces a segmentation fault on exit.

**To Reproduce**
1. Try to compute vascular segments without having the ExtractCenterline module loaded/installed

**Expected behavior**
2. An error should warn the user that the the module ExtractCenterline is missing.
3. The application should continue running as normal
4. No segmentation faults should be produced on termination

**Desktop (please complete the following information):**
 - OS: GNU/Linux Gentoo
 - Version Slicer-master (5.3)
 
**Additional context**
This happens under a development environment where everything is built from source. I theory, this should never happen in pre-built binaries.

gatangen:


olevs (approved):

RafaelPalomar:
Hi, @gatangen this PR has still conflicts (the usual suspect Liver.py)

Issue #206, opened 2023-03-01
Better defaults for initial placment of resection markups


RafaelPalomar:
Right now the curved initialization of resection is not very useful, we need to come up with better resection initialization.

Suggestion: Using the camaera parameters  from tumor (or center of parenchyma) to the right (or left) out of the bounding box of the parenchyma

PR #247, opened 2023-08-22
BUG: Fix defition of initial position and preprocessing

### LINKED ISSUE #242: Return of different data types based on previous condition
We should change this fragment so the function in question does not return different data type (`vtkPolyData` vs. `vtkMRMLNode`). Caller functions of this do not handle different data types, but assume the return is `vtkMRMLNode`.

https://github.com/ALive-research/Slicer-Liver/blob/40402d63fd223cf59be39954c5f620c8984f0199/Liver/Liver.py#L1092-L1094

dalbenzioG:
This should close #242 and #206


RafaelPalomar:
Is this ready for review?

RafaelPalomar (approved):

PR #255, opened 2023-09-25
ENH: fixed #248, adjusted region growing and surface resolution settings, new function for export selected segments as new segmentation node

### LINKED ISSUE #248: The data type is not fixed in the liver volumetry module segment selection
**Describe the bug**
As the title said, when trying to calculate the selected volume without placing markers and resections, the error happens.

RuoyanMeng:
#266 #248


RafaelPalomar:
@RuoyanMeng, could you briefly describe what are the bug and the fix?

LiverVolumetry/LiverVolumetry.py
```diff
@@ -387,8 +387,14 @@ def computeVolume(self, segmentsVolumeNode, targetSegmentVolumeNode, segmentatio
...
-          voxel_count = stats[segmentId,"LabelmapSegmentStatisticsPlugin.voxel_count"]
-          volume_cm3 = stats[segmentId,"LabelmapSegmentStatisticsPlugin.volume_cm3"]
+          voxel_count = 0
+          volume_cm3 = 0
+          if stats[segmentId,"LabelmapSegmentStatisticsPlugin.voxel_count"]:
+            voxel_count = stats[segmentId,"LabelmapSegmentStatisticsPlugin.voxel_count"]
+            volume_cm3 = stats[segmentId,"LabelmapSegmentStatisticsPlugin.volume_cm3"]
+          elif stats[segmentId,"ScalarVolumeSegmentStatisticsPlugin.voxel_count"]:
+            voxel_count = stats[segmentId,"ScalarVolumeSegmentStatisticsPlugin.voxel_count"]
+            volume_cm3 = stats[segmentId,"ScalarVolumeSegmentStatisticsPlugin.volume_cm3"]
```
RafaelPalomar, line 397: Please explain this snippet. Why the need to have LabelMap and ScalarVolume counts?

RuoyanMeng: Although I added the selected segmentations combined node as a "vtkMRMLLabelMapVolumeNode", in some cases, it happens that the input turns into ScalarVolume data and pops up error messages about couldn't find LabelmapSegmentStatisticsPlugin. But lately, I haven't met this issue, just in case it would happen to a different dataset, I added those scripts to ensure that the function works.

RuoyanMeng: Region growing setting changes because we have more than 7 labels in the original data, to avoid conflicts, adjust the base value to 100.

RuoyanMeng: Resolution setting change is a temporary solution before I develop a stable method. For most cases, it needs at least 500*500 to ensure successful region growing for FaceConnectivity.

RafaelPalomar (approved):

PR #282, opened 2025-04-08
BUG: Fixed when reopening saved vascular territory segmentation segments name list was incorrectly read.

RuoyanMeng:
This function updateVascTerrList intends to fetch all the segments' names. However, in the code, it reads segmentID which is different from the segment Name leading to the unsuccessful recovery of vascular territory segmentation info.

### DIFF
--- a/LiverSegments/LiverSegments.py
+++ b/LiverSegments/LiverSegments.py
@@ -321,21 +321,26 @@ class LiverSegmentsWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
 
   def updateVascTerrList(self, vasc_terr_ID_list, vascular_territory_segm_node):
-    segments = vascular_territory_segm_node.GetSegmentation().GetSegmentIDs()
+    segmentNames = []
+    segmentIds = vascular_territory_segm_node.GetSegmentation().GetSegmentIDs()
+    for id in segmentIds:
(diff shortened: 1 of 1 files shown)


olevs (approved):

LiverSegments/LiverSegments.py
```diff
@@ -321,18 +321,23 @@ def vascular_territory_segmentationNodeSelected(self):
...
     vasc_terr_ID_list.addItem(initString)
     firstSegmentID = 'Vascular Territory ID 1'
-    if firstSegmentID not in segments:
+    if firstSegmentID not in segmentNames:
       # No vascular territory segmentations
       return
     #Start populating Vascular Territory list
     index = 0
-    for idString in segments:
+    for idString in segmentNames:
```
RafaelPalomar, line 340: @RuoyanMeng. Mixing `ID` and `name` can be confusing in the code due to the fact that MRML nodes ha `GetID` and `GetName`, which are tow different concepts treated differently. Please, make sure that variables with `id` refere to an actual id; similarly to variables with `name`

RafaelPalomar (changes requested):

LiverSegments/LiverSegments.py
RuoyanMeng: Hi Rafael, Thanks for the suggestion. I updated the code naming, and hopefully it's clearer now.

PR #288, opened 2026-05-13
DOC: Introduce architecture diagrams + ADR scaffolding

RafaelPalomar:
## Summary

Bootstraps an **architectural intent layer** for Slicer-Liver as committed, diffable text:

- `Docs/architecture/` — PlantUML / Mermaid diagrams (`current-*` vs `target-*` naming) that capture MRML hierarchy, module dependencies, observer/event flow, and external integration boundaries.
- `Docs/adr/` — Nygard-style Architecture Decision Records, append-only, with a numbered template.

This is **scaffolding only** — no diagrams or ADRs are populated yet. The next PR will land `target-mrml-node-hierarchy.puml` plus an ADR justifying the `vtkMRMLLiverResectionNode` shape, after which the scaffolding starts earning its keep.

The intent is for these artefacts to act as the spec that future PRs (especially the planned upcoming refactor work) are reviewed against — drift from a committed `target-*` diagram or contradiction of an `Accepted` ADR should produce review comments with the relevant artefact cited.

## Test plan

- [x] Markdown renders cleanly on GitHub
- [x] Commit messages follow Slicer-Liver convention (`DOC:`, `STYLE:`)
- [x] Validated end-to-end via local multi-agent review (architecture-conformance, MRML/VTK correctness, style, test coverage); style reviewer caught three personal-environment leaks which are fixed in `a1132d2`
- [x] Reviewer approval that the convention is workable before populating the first real diagram + ADR

---


Docs/architecture/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Naming convention
+
+Diagrams come in two flavours, distinguished by filename prefix:
+
+| Prefix | Meaning | Example |
+|---|---|---|
+| `current-` | The code as it is **today**.  Snapshot, evolves slowly. | `current-mrml-node-hierarchy.puml` |
+| `target-` | The architecture the refactor is moving **toward**. | `target-mrml-node-hierarchy.puml` |
+
```
RafaelPalomar: Didn't we agreed this would be mermaid diagrams?

Docs/architecture/README.md
```diff
@@ -0,0 +1,92 @@
...
+The `/slicer-review` command reads every file in this directory into the
+review context and grades changed code against it:
+
+- A PR that introduces a new `vtkMRMLLiverFooNode` without updating
+  `target-mrml-node-hierarchy.puml` triggers a **drift** comment.
+- A PR that matches a `target-*` diagram triggers a positive **advances**
+  comment.
+- A change that touches code referenced by a `current-*` diagram should
+  update that diagram in the same PR (or explicitly justify in the ADR
+  what changed).
```
RafaelPalomar: This needs clarification. /slicer-review is a command for AI agent and it is not mandatory:
  - Should we include the skill in this repo? Otherwise where is the skill found (it looks more generic skill than just Slicer-Liver?

Docs/architecture/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+- **Browser-based**: paste the fence content into
+  https://mermaid.live for an immediate preview.
+- **Editor plugins**: VS Code (`bierner.markdown-mermaid`), Emacs
+  (`ob-mermaid`), IntelliJ (`mermaid` plugin).
+- **CLI**: `mmdc` (mermaid-cli) for export to SVG/PNG when needed —
+  rarely necessary now that GitHub renders fences.
+
+For the (rare) PlantUML case, install `plantuml` from your package
+manager of choice and render with `plantuml -tsvg path/to/file.puml`.
```
RafaelPalomar: With mermaid should be enough. I don't think we should keep plantuml as an option

Docs/architecture/README.md
RafaelPalomar: Fixed in 3d1a96a: naming-convention table examples now use `.md` extensions (`current-mrml-node-hierarchy.md` and `target-mrml-node-hierarchy.md`), matching the Mermaid-only policy. Also dropped the PlantUML mentions elsewhere in the doc per the L86 comment.

Docs/architecture/README.md
RafaelPalomar: Rewritten in 3d1a96a. The section now makes three things explicit:

1. The rules of thumb (drift / advances / update-the-diagram) stand on their own for any human reviewer — no tooling dependency.
2. `/slicer-review` is an **optional** AI-agent slash-command the project lead uses; it is not a project-wide gate, and contributors are not expected to install or run it.
3. The skill is generic enough to apply to other Slicer extensions and is therefore **not vendored** in this repo; it lives in the project lead's personal AI-agent configuration. Packaging for reuse is flagged as a future concern.

Docs/architecture/README.md
RafaelPalomar: Done in 3d1a96a: PlantUML removed entirely from the document — the "What lives here" section now reads "every diagram is a fenced Mermaid block" with no fallback option, and the rendering / authoring tooling section drops the PlantUML CLI note.

Docs/architecture/README.md
RafaelPalomar: Follow-up: the skill has since been published as a standalone repo at https://github.com/OUH-MESHLab/slicer-review (BSD-3-Clause, public). The README in this PR (commit e341e63) was updated to link directly to that repo rather than describing it as "lives in the project lead's personal AI-agent configuration." The newer wording:

> An **optional** Claude Code slash-command, [`/slicer-review`](https://github.com/OUH-MESHLab/slicer-review), reads this directory into review context and applies the same rules of thumb (plus Slicer-specific MRML/VTK correctness checks). It is one reviewer's tool, **not a project-wide gate** […]
>
> The skill is maintained as a separate, reusable repository under the [OUH-MESHLab](https://github.com/OUH-MESHLab) organisation rather than vendored here, since it applies generically across 3D Slicer extension repositories.

So the "where does the skill live" question now has a public answer.

RafaelPalomar (commented):
TLGTM

PR #289, opened 2026-05-13
DOC: Current MRML node hierarchy + ADR-0001 (three-node resection)

RafaelPalomar:
> Stacked on #288 — review only after the scaffold convention is settled.

## Summary

Lands the first concrete artefacts on top of the architecture scaffolding from #288, making the spec layer **load-bearing** for subsequent code review:

- **`Docs/architecture/current-mrml-node-hierarchy.puml`** (+ rendered `.svg`) — class diagram of the current MRML node graph for `LiverResections` + `LiverMarkups` modules, showing the three-node resection assembly, storage-node delegation, and the 2D displayable manager.
- **`Docs/adr/0001-resection-three-node-assembly.md`** — Nygard-style ADR recording **why** a resection is split across three sibling nodes (Initialization contour + Bezier geometry + Resection content) rather than a single content node. The split is a preference for clean serialization, not an impossibility — see Context and rejected Alternative B for the actual force.

## Workflow validation

Ran `/slicer-review` locally on commit `def6d77` before pushing. Four parallel reviewers:

| Reviewer | Verdict | Notable finding |
|---|---|---|
| Architecture conformance | advance, spec layer "loaded" | none |
| MRML/VTK correctness (fact-check) | **2 factual errors, both fixed in `7c94cea`** | ADR claimed MarkupsNode is not Storable (it is, transitively via DisplayableNode → TransformableNode → StorableNode); commit range cited as 2019 was actually empty and the wrong years (real introducing commits are 2021-07-01) |
| Style conformance | clean | commit subject 76 chars, minor |
| Test coverage | N/A direct; **gap flagged** | no test covers `.lrp.fcsv` save/reload — this is the safety-net characterization test that must exist *before* any of the cleanup ADRs (0002, 0003) referenced in 0001's Consequences can be safely attempted |

The factual-error catch is exactly the loop's value moment: the ADR would have committed wrong technical claims as institutional fact had the reviewer not been invoked.

## Test plan

- [x] PlantUML renders cleanly (rendered SVG committed, 57 KB)
- [x] ADR template fields all present (Status, Date, Deciders, Diagrams, PR, Context, Decision, Alternatives, Consequences, References)
- [x] Internal links resolve
- [x] Class names verified against `LiverResections/MRML/`, `LiverMarkups/MRML/`, `Slicer/Libs/MRML/Core/`
- [x] Inheritance chain in diagram verified against actual Slicer source
[...]

PR #290, opened 2026-05-13
DOC: ADR-0001 status amendment (descriptive only; see ADR-0002)

RafaelPalomar:
> Stacked on #289 — read after #288 (scaffolding) and #289 (current-state diagram + ADR-0001) have been reviewed.

## Summary

Adds an Update block at the top of ADR-0001 and changes its Status to **Accepted (descriptive — records current state and historical rationale; superseded by ADR-0002 for target architecture)**.

The body below the Update is preserved unchanged — the original four-field Nygard rationale stands as the historical record of why the code is shaped as it is today.  Append-only ADR discipline.

## Why now

The pending refactor direction is **migration to SlicerLayerDisplayableManager (LayerDM)**, which dissolves the \`vtkMRMLMarkupsNode\` inheritance constraint that originally forced the three-node resection assembly.  Without this amendment, ADR-0001 reads as an endorsement of the current Markups-based design, misleading future contributors into thinking the three-node split is the desired end-state.

## Stacked ADRs that follow this one

| ADR | Topic | PR |
|---|---|---|
| 0002 | Migrate to SlicerLayerDisplayableManager (target) | next in stack |
| 0003 | Testability invariant (every PR pins behaviour) | next in stack |
| 0004 | Python by default; C++ only for nodes + algorithms | next in stack |

## Test plan

- [x] ADR template fields all present after amendment
- [x] Internal link to ADR-0002 resolves once the next PR lands
- [ ] Reviewer agreement that "descriptive" is the right framing (vs full supersession + new replacement ADR)

---

### DIFF
--- a/Docs/adr/0001-resection-three-node-assembly.md
+++ b/Docs/adr/0001-resection-three-node-assembly.md
@@ -1,10 +1,25 @@
 # 0001. Resection is a three-node MRML assembly
 
-- **Status:** Proposed
+- **Status:** Accepted (descriptive — records the current state and its
+  historical rationale; superseded by [ADR-0002](0002-migrate-to-slicerlayerdm.md)
+  for target architecture)
 - **Date:** 2026-05-13
 - **Deciders:** Rafael Palomar (recording rationale on behalf of original authors)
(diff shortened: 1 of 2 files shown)


RafaelPalomar (commented):
LGTM

PR #293, opened 2026-05-13
DOC: ADR-0004 — Python by default, C++ for nodes + algorithms

RafaelPalomar:
> Stacked on the ADR-0003 testability-invariant PR.  This sets the **implementation-language boundary** for everything that follows.

## Summary

The C++/Python boundary for Slicer-Liver:

**C++ only for:**
1. MRML node classes, kept **data-only** (no business logic, no display state)
2. Algorithm libraries (Bezier eval, distance maps, contour interpolation, volumetry, custom VTK filters)

**Python for everything else:**
- Module / Widget / Logic classes (\`ScriptedLoadableModule\` pattern)
- Storage I/O (\`.lrp.fcsv\` reader/writer)
- LayerDM pipelines (per ADR-0002)
- Tests (per ADR-0003)

## Why this shape

- LayerDM's first-class Python pipeline API erases the widget-pipeline reason that pushed historical Slicer extensions toward all-C++.
- Refactor cost dominates over a multi-year migration horizon — Python edits are seconds, C++ is a Slicer rebuild.
- Researcher-heavy contributor base; lowering the barrier without giving up performance requires shrinking the C++ surface to the narrow band where C++ pays back.
- The testability invariant (ADR-0003) is dramatically cheaper to honour in Python.

## Rejected alternatives (full reasoning in ADR)

- **All C++** — maintenance burden, contributor barrier, wastes LayerDM's Python API.
- **All Python including MRML nodes via \`vtkMRMLScriptedNode\`** — escape hatch too limited for clinical artefacts (weaker scene introspection, rougher storage round-tripping, wrong risk surface).
- **Case-by-case without explicit boundary** — every PR re-litigates the same question.
- **Python everywhere including algorithms** — existing C++ tuned for inner loops, profiling cost outweighs gain.

## Test plan

- [x] Reviewer agrees \`vtkMRMLScriptedNode\` is genuinely too limited for clinical use (vs the data-only C++ MRML node pattern)
- [x] Reviewer agrees the algorithm library boundary is drawn at the right place
- [x] Reviewer agrees the \`/slicer-review\` reviewer can enforce this convention going forward

---

### DIFF
--- a/Docs/adr/0004-python-cpp-boundary.md
+++ b/Docs/adr/0004-python-cpp-boundary.md
@@ -24,7 +24,7 @@ Two reasons historically pushed Slicer extensions toward C++:
    resolution** are conventionally done by C++ subclasses with proper
    `Set/Get` accessors, `Modified()` events, and `Copy()`/
    `PrintSelf()` / `WriteXML()` / `ReadXMLAttributes()` semantics.
-   The `vtkMRMLScriptedNode` mechanism exists for declaring node
+   The `vtkMRMLScriptedModuleNode` mechanism exists for declaring node
    attributes from Python but is limited compared to a proper C++
    subclass — and the limitations are precisely the ones that hurt
    for a clinical artefact (introspectability, storage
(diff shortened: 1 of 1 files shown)


RafaelPalomar (commented):
LGTM

PR #294, opened 2026-05-13
ENH: Characterisation test for vtkMRMLLiverResectionCSVStorageNode round-trip

RafaelPalomar:
> Stacked on #293 (ADR-0004). Per ADR-0003, this is the **safety net** that must land (and be verified against a real build) before any LayerDM-migration code touches the storage path.

## Summary

Adds `vtkMRMLLiverResectionStorageRoundTripTest` — a characterisation test that pins today's `.lrp.fcsv` round-trip behaviour of `vtkMRMLLiverResectionCSVStorageNode`.

### What it pins

| Phase | Step | Assertion |
|---|---|---|
| 1 | Build resection + Bezier assembly; populate 16 deterministic control points; `WriteData` | `WriteData` returns 1 |
| 2 | `scene.Clear()`; rebuild fresh assembly | New Bezier has 0 control points before read |
| 3 | `ReadData` into the fresh assembly | All 16 points survive coordinate-by-coordinate within 1e-6 |
| 4 | Probe failure modes | `WriteData` fails when no Bezier ref; `CanReadInReferenceNode` accepts resection, rejects bare Bezier |

### Why now

ADR-0001's Consequences section explicitly identifies the `.lrp.fcsv` reload-ordering as the fragile path. ADR-0003 makes a characterisation test for this path the precondition for any of the LayerDM-migration PRs (ADR-0002).

Without this test, the LayerDM migration would refactor a storage path with no behaviour pin — any regression would surface only when a surgeon tries to reload a planning file in clinical use, which is exactly the wrong feedback loop.

### Language choice

C++, matching the existing `LiverResections/Testing/Cxx/` pattern. ADR-0003 permits both C++ and Python; ADR-0004 prefers Python by default but for retroactive characterisation of an existing C++ module the existing pattern wins.  When LayerDM-era Python test infrastructure lands, new tests should follow ADR-0004; this one stays C++ as a stable safety net.

## Test plan

- [ ] **Build against a real Slicer tree and run** — the test has not been validated on hardware.  Any divergence from the actual API surfaces at compile time first.
- [ ] All four phases pass on a clean build of `refactor/characterization-test-resection-roundtrip` on top of `develop`.
- [ ] Reviewer confirms the four-phase structure pins the right contract (not just a tautology).
- [ ] Reviewer confirms the failure-mode probes in Phase 4 are not too tight (they should pin the storage class's defensive checks, not the exact error messages).

## Known caveats

- The test uses `RemoveAllControlPoints` + `AddControlPoint` on `vtkMRMLMarkupsBezierSurfaceNode`; the Bezier node's `RequiredNumberOfControlPoints = 16` constraint may interact with this in ways that need verification on a real build.
- `TESTING_OUTPUT_IGNORE_WARNINGS_ERRORS_BEGIN/END` is used around the Phase-4 no-Bezier-ref probe to silence the expected `vtkErrorMacro` from `WriteDataInternal` so `WITH_VTK_ERROR_OUTPUT_CHECK` does not flag it as unexpected.

---


RafaelPalomar (commented):
LGTM

PR #317, opened 2026-05-15
DOC: workflow diagrams for current-state Liver modules (T1-F)

RafaelPalomar:
## Summary

Adds Mermaid workflow diagrams for the three end-to-end Liver modules
under a new `Docs/architecture/ui/` directory, per ADR-0009 §2.  These
are the **current-state** baseline — the *before* picture that
ADR-0009-compliant PRs amend from v2.0.0 onward.

Each diagram file contains:

1. One opening paragraph on what the module does for the user.
2. A `flowchart TD` of the user's path (clicks, panel order, gates).
3. A `stateDiagram-v2` where the module exposes interactive state
   (LiverResections only; the other two are gate-driven push-button
   flows and are flagged as such).
4. A `Known gaps / pain points` bullet list — observations only, no
   recommendations.

## Modules covered

- `Docs/architecture/ui/liver-resections.md` — the resection-planning
  surface hosted in the Liver scripted module
  (DistanceMaps + Resections + Resectogram sub-panels), plus the
  `ResectionState` / `InitializationMode` enum-driven state machine
  on `vtkMRMLLiverResectionNode`.
- `Docs/architecture/ui/liver-segments.md` — the
  vascular-territory-segmentation workflow (hidden module surfaced
  inside Liver), with a coarse Disabled / Idle / Placing /
  CenterlineReady / Computed state diagram derived from the widget's
  gating behaviour.
- `Docs/architecture/ui/liver-volumetry.md` — the resection-volumetry
  workflow (hidden module surfaced inside Liver).  No multi-state
  widget present; the state-machine section is intentionally
  minimal per ADR-0009 §1.

## Out of scope

- No source-code edits.  Only files under `Docs/architecture/ui/`
  touched.
- Observations are surfaced as "Known gaps / pain points" without
  recommendations or fixes — they exist to inform later
  ADR-0009-compliant PRs.

Closes #312.

---

*AI-assisted authorship: this pull request was drafted with help from Anthropic's Claude (Opus 4.7, `claude-opus-4-7`) via Claude Code. Plan and brief authored by the orchestrating Opus 4.7 session; workflow diagrams drafted by a Claude Code subagent under that plan. Opened for human review before merge.*

PR #316, opened 2026-05-15
ENH: pytest scaffold with render_interactive option per ADR-0008

RafaelPalomar:
## Summary

- Lands the canonical pytest scaffold described in [ADR-0008](Docs/adr/0008-testing-strategy.md): shared `Testing/Python/` tree, `conftest.py`, `--render-interactive` CLI option, and the session-scoped `render_interactive` fixture.
- Adopts the trame-slicer / SlicerLayerDisplayableManager dual-use pattern — the same test code serves CI regression (offscreen) and developer interactive exploration (`pytest --render-interactive=5`).
- Two CTest entries (`pytest_scaffold` and `pytest_scaffold_render_interactive`) mirror the CI matrix in ADR-0008 §6.

## Default for `--render-interactive`

Default is **`0.0`** (offscreen).  Rationale:

- ADR-0008 §6 specifies that CI's brief-onscreen pass is an explicit *separate* invocation (`pytest --render-interactive=0.1`).  Bare `pytest` must therefore default to offscreen so the matrix's two invocations exercise distinct paths.
- GitHub Actions runners drive Qt via `QT_QPA_PLATFORM=offscreen` (`.github/workflows/ci.yml`); a default-on flag would hang on missing DISPLAY or burn wall-clock on invisible windows.
- Matches trame-slicer's convention.

## Implemented

- `Testing/Python/conftest.py` — `pytest_addoption`, `render_interactive` fixture (session-scoped, returns float seconds), `a_slicer_app` example fixture demonstrating the Slicer-imported-as-library pattern with graceful skip when Slicer is absent.
- `Testing/Python/unit/test_scaffold_smoke.py` — three unit-layer tests pinning the option's registered default (`0.0`), the fixture's contract (non-negative float), and Slicer-independence of the unit layer.
- `Testing/Python/workflow/test_render_interactive_dispatch.py` — two workflow-layer tests that exercise the dispatch pattern in both modes (no real view yet — that fixture lands in a follow-up).
- `pytest.ini` — `testpaths`, `python_files = test_*.py` (deliberately excludes the existing Slicer-self-test `*Test.py` files so they continue to run through `slicer_add_python_unittest` without colliding).
- `Testing/CMakeLists.txt` + `Testing/Python/CMakeLists.txt` — CTest registration gated on `BUILD_TESTING` and on `pytest` being importable from the configured Python.
- `CMakeLists.txt` top-level — `add_subdirectory(Testing)`.
[...]

PR #315, opened 2026-05-15
ENH: SNOMED-CT terminology assets and per-tool LabelToSCT bridges (ADR-0011)

RafaelPalomar:
## Summary
- `Resources/Terminology/SlicerLiver-Terminology.json` — dcmqi segment-context schema, three categories: Anatomical Structure (Liver, Portal vein, Hepatic vein), Morphologically Altered Structure (Neoplasm Primary), Liver segment (private `99SLIVER` scheme, ten Couinaud codes).
- `Resources/Terminology/LabelToSCT/{TotalSegmentator,MONAILabel,KumarOram}.json` — per-tool label-to-SCT bridges per ADR-0011.

Closes #309.

## SCT code choices

| Concept | Code | Note |
|---|---|---|
| Liver | `SCT:10200004` | Matches Slicer DICOM-Master |
| Portal vein | `SCT:32764006` | Matches Slicer DICOM-Master |
| Hepatic vein | `SCT:8993003` | Matches Slicer DICOM-Master |
| Tumor | `SCT:86049000` "Neoplasm, Primary" | DICOM SEG / MONAI ecosystem default for hepatic tumor segmentation |

The portal/hepatic vein codes here intentionally diverge from a widely-cited alternate pair (`12830003` / `80334008`) — those are not the values Slicer ships in its bundled DICOM-Master terminology and would break interop with the Segment Editor selector and DICOM SEG export. Verified against `SegmentationCategoryTypeModifier-DICOM-Master.json` in upstream Slicer 5.8.1.

## Couinaud scheme

Ten codes in private scheme `99SLIVER`:
- `LIVER-SEG-I` through `LIVER-SEG-VIII` — standard eight Couinaud segments
- `LIVER-HEMI-RIGHT` (Couinaud V-VIII) and `LIVER-HEMI-LEFT` (Couinaud II-IV) — compound regions for anatomic-resection workflows

Per ADR-0011 §Versioning, the eight standard codes and the two hemiliver codes are immutable identifiers for v2.x. Adding finer subdivisions (e.g. IVa/IVb) is MINOR; renumbering or removing existing codes is MAJOR (N-class break under ADR-0007).

## Stubs / open items

- **TotalSegmentator** — `inferior_vena_cava` and `hepatic_vessels` are marked out-of-scope (Slicer-Liver core type set doesn't cover them in v2.0.0). They pass through as Unknown.
[...]

PR #324, opened 2026-05-15
DOC: ADR-0013 — LayerDM Pipeline pattern for Slicer-Liver

RafaelPalomar:
## Summary
Commits ADR-0013 (LayerDM Pipeline pattern) — previously drafted during a PKS-side design discussion and left uncommitted in the preview worktree per `.claude/handoff-2026-05-15-v2-architecture.md`.

The ADR fixes the *shape* every Slicer-Liver LayerDM migration instantiates: one Python Pipeline class per display-node type, C++ data-only display nodes, state-aware Pipelines observing both the display node and orchestrating-state nodes, composable Representations, and the three-tier test set per ADR-0008.

## Migration map (per ADR-0012 as amended)
Two phases:
1. **T2 — LiverResections (all-in)** absorbing the LiverMarkups dissolution per the forthcoming ADR-0014, with C++ algorithm lifts per the forthcoming ADR-0015.
2. **T3 — Resectogram + distance maps**.

ADR-0012 defers LiverSegments / LiverVolumetry / Modeling LayerDM display nodes to v2.1.0; they instantiate the same pattern when they land.

## Key decisions captured
- §1 One Python Pipeline class per display-node type (not per data node — preserves the compositional axis)
- §2 C++ side stays data-only per ADR-0004
- §3 Pipeline responsibilities (observe display node, own Representations, dispatch SCT terminology, idempotent `update()`)
- §4 Pipelines may observe orchestrating-state nodes — supports ADR-0014's three state-conditional Representations
- §5 Lifecycle via `vtkMRMLLayerDMPipelineManager`; Pipelines not persisted
- §6 Representations as composable testable units (one VTK actor + mappers; small Python class)
- §7 Naming and file layout
- §8 Data nodes shrink — display state leaves the data node (D-class break already on the v1→v2 ticket)
- §9 Three-tier test set per ADR-0008 (module / unit / workflow with `render_interactive`)
- §10 Illustrative Pipeline class skeleton

## Cross-refs
- ADR-0002 (LayerDM as the framework target)
- ADR-0004 (Python/C++ language boundary that puts Pipelines in Python)
- ADR-0008 (three-layer test discipline; the real-view fixture deferred by PR #316 lands in T2)
- ADR-0011 (SCT dispatch happens inside the Pipeline)
- ADR-0012 (the v2.0.0 scope this pattern implements; companion PR amends 0012 to the two-phase map)
- ADR-0014 + 0015 (forthcoming this week — the T2 implementation specifics)

## UX impact
[...]

PR #326, opened 2026-05-15
DOC: ADR-0014 + ADR-0015 — LiverMarkups dissolution and C++ algorithm library

RafaelPalomar:
## Summary

Two co-dependent ADRs landing together to fix the v2.0.0 architecture for the
resection-planning workflow.  Both `Status: Proposed`; status flips to
`Accepted` on merge per `Docs/adr/README.md`.

### ADR-0014 — Dissolve LiverMarkups; fold interactive resection primitives into LiverResections

The `LiverMarkups` module disappears entirely in v2.0.0.  Its three
Markups-derived node pairs (`vtkMRMLMarkupsBezierSurfaceNode`,
`vtkMRMLMarkupsSlicingContourNode`, `vtkMRMLMarkupsDistanceContourNode`)
relocate into `LiverResections` as **non-Markups data nodes**, driven by a
single state-aware LayerDM Pipeline (per [ADR-0013](../blob/preview/Docs/adr/0013-layerdm-pipeline-pattern.md) §4)
that owns three state-conditional Representations:

- `SlicingPlaneInitRepresentation` when `(ResectionState=Init, InitializationMode=SlicingPlane)`,
- `DistanceSpheroidInitRepresentation` when `(ResectionState=Init, InitializationMode=DistanceSpheroid)`,
- `BezierPlanningRepresentation` when `(ResectionState=Planning, *)`.

The reframing is that `SlicingContour` and `DistanceContour` are **alternative
initialisation affordances** for the same Bezier surface — not independent
annotations.  Two forces push the Bezier surface off `vtkMRMLMarkupsNode`:
ring-aware right-click on the 4×4 control grid (interior / edge / corner
groups), and per-role glyph rendering through the existing
`vtkOpenGLBezierResectionPolyDataMapper` family in `LiverMarkups/VTKWidgets/`.
The dissolution lets a single `vtkLiverBezierWidget` subclass `vtkAbstractWidget`
directly with a clean left-drag / right-drag / right-click event table, and
makes the state machine first-class on `vtkMRMLLiverResectionNode`
(`ResectionState`, `InitializationMode`) instead of implicit in scene
contents.

Init control points persist across the Init→Planning transition (recoverable
Planning→Init drop-back).  Storage rides through `LiverResections`'
`.lrp.fcsv` with a schema bump.  This is a **D-class + N-class** break per
[ADR-0007](../blob/preview/Docs/adr/0007-version-numbering-policy.md), already
on the v1→v2 ticket.

### ADR-0015 — Lift parameterisation and fitting algorithms from Python to a C++ algorithm library under LiverResections

The Init→Planning transition's Bezier fit lifts out of three suspected-buggy
Python functions in `Liver/Liver.py` (`fit_bezier_surface`,
`runSurfacefromCurve`, `runSurfacefromEFD`) into a new
`LiverResections/Algorithm/` subdirectory housing four
`vtkPolyDataAlgorithm`-style classes:

| Class | Used by |
|---|---|
| `vtkLiverPlaneRingExtractor` | `SlicingPlaneInitRepresentation` |
| `vtkLiverSpheroidRingExtractor` | `DistanceSpheroidInitRepresentation` |
| `vtkLiverContourParameterizer` | `vtkLiverBezierFitter` |
| `vtkLiverBezierFitter` | Init→Planning transition |

Pure VTK, **no MRML dependency**, Python-wrapped via standard VTK wrapping,
tested under `Algorithm/Testing/Cxx/` with no Slicer import (the
[ADR-0008](../blob/preview/Docs/adr/0008-testing-strategy.md) C++ low-level
test layer — `ctest -R BezierFitter` is a one-line bug-repro).  Migration of
the three Python functions follows [ADR-0003](../blob/preview/Docs/adr/0003-testability-invariant.md)
characterisation discipline: characterisation tests on current Python
behaviour land first, then the lift inverts the assertion for bit-equivalence
(or documented numerical tolerance, e.g. on the EFD path).

This is the **first concrete realisation of [ADR-0004](../blob/preview/Docs/adr/0004-python-cpp-boundary.md)**'s
Python/C++ boundary; v2.1.0 work (volumetry, distance maps) follows the
same in-tree pattern.

## Cross-references

- **[ADR-0013](../blob/preview/Docs/adr/0013-layerdm-pipeline-pattern.md)
  §4 (state-aware Pipelines)** — ADR-0014 is its first concrete
  instantiation; the three Representations are owned by one Pipeline
  observing both its display node and `vtkMRMLLiverResectionNode`'s
  state-machine enums.
- **[ADR-0012](../blob/preview/Docs/adr/0012-layerdm-migration-v2-scope.md)
  amendment (sibling PR)** — the v2.0.0 migration map collapses to
  **T2 LiverResections all-in + T3 Resectogram**.  This PR documents the
  decision; the ADR-0012 amendment PR records it in the scope ADR.
- **[ADR-0011](../blob/preview/Docs/adr/0011-sct-terminology-dispatch.md)** —
  the new `vtkMRMLLiverBezierSurfaceDisplayNode` consumes SCT triples for
  colour / label decoration, replacing the hardcoded `HepaticContourColor` /
  `PortalContourColor` constants in the current `LiverMarkups` node
  constructors.
- **[ADR-0007](../blob/preview/Docs/adr/0007-version-numbering-policy.md)** —
  ADR-0014 carries the (D) and (N) MAJOR triggers under the v1→v2 jump;
  ADR-0015 deliberately avoids an (E) trigger by hosting the algorithms
  in-tree rather than depending on CGAL/libigl.
- **[ADR-0008](../blob/preview/Docs/adr/0008-testing-strategy.md)** —
  ADR-0014's Pipeline ships the three-tier test set (module +
  Representations + workflow with `render_interactive`); ADR-0015's library
  ships the C++ low-level test set.
- **PR #294** (characterisation harness) — extended by ADR-0014 step 1.
- **PR #315** (SCT terminology assets) — consumed by the new display node.
- **PR #316** (pytest scaffold) — hosts the Python wrapper tests for the
  ADR-0015 algorithm library.
- **PR #317** (workflow-diagram baseline) — `Docs/architecture/ui/liver-resections.md`
  refreshes against the state-aware Pipeline.

## Verification

- [x] Both files match `Docs/adr/0000-template.md` structure (Status,
      Date, Deciders, Diagrams, PR, Context, Decision, Alternatives
      considered, Consequences, References).
- [x] All forward references resolve (ADR-0014 ↔ ADR-0015 cross-link
      cleanly; both forward-reference the ADR-0012 amendment landing in a
      sibling PR and the ADR-0013 §4 amendment).
- [x] Code block in ADR-0014 §6 stays at 13 lines (under the ≤15-line
      ceiling).
- [x] ADR-0014 ships 3 Alternatives (A/B/C) and 4 Consequences as the
      brief specifies; ADR-0015 ships 4 Alternatives (A/B/C/D) and 4
      Consequences.
- [x] No source-code or build-system changes touched; this PR is
      `Docs/adr/` only.

## Test plan

- [ ] Human review of the two ADRs for prose register, decision
      precision, and alternative-rejection rationale.
- [ ] Confirm sibling ADR-0012 amendment and ADR-0013 §4 amendment PRs
      land before this one (or co-stack cleanly) so the cross-references
      resolve to `Accepted` ADRs on merge.
- [ ] On merge: flip both ADRs to `Status: Accepted` and fill the PR link
      per `Docs/adr/README.md`'s lifecycle convention.

---

*AI-assisted authorship: this pull request was drafted with help from Anthropic's Claude (Opus 4.7, `claude-opus-4-7`) via Claude Code. Plan and brief from the orchestrating Opus 4.7 session per the handoff at `.claude/handoff-2026-05-15-v2-architecture.md`; prose drafted by a Claude Code subagent. Opened for human review before merge.*

### DIFF
--- a/Docs/adr/0014-livermarkups-dissolution.md
+++ b/Docs/adr/0014-livermarkups-dissolution.md
@@ -69,7 +69,7 @@ lived **implicitly in scene contents**: "if a `SlicingContour` node
 exists, we are in Initialization with SlicingPlane mode; if a
 `DistanceContour` node exists, we are in Initialization with
 DistanceSpheroid mode; if a `BezierSurface` node exists, we are in
-Planning".  [ADR-0013](0013-layerdm-pipeline-pattern.md) §4 documents
+Planning".  ADR-0013 §4 documents
 the brittleness of that pattern: ambiguous scene round-trip mid-
 transition, undo/redo crossing node boundaries, every state check is a
 scene query.  PR #317's UI baseline at
@@ -85,7 +85,7 @@ scene-contents pattern altogether.
 The `LiverMarkups` module is dissolved in v2.0.0.  Its three
 interactive primitives relocate into `LiverResections` as **non-Markups
 data nodes**, driven by a single state-aware LayerDM Pipeline with
-three Representations (per [ADR-0013](0013-layerdm-pipeline-pattern.md)
+three Representations (per ADR-0013
 §4 and §6) and a single custom widget that subclasses
 `vtkAbstractWidget` directly.
 
@@ -115,7 +115,7 @@ Planning→Init drop-back recovers them losslessly.
 
 ### 2. One LayerDM Pipeline; three state-conditional Representations
 
-Per [ADR-0013](0013-layerdm-pipeline-pattern.md) §4, a single
+Per ADR-0013 §4, a single
 `LiverBezierSurfacePipeline` observes its display node *and* the
 parent `vtkMRMLLiverResectionNode`'s `ResectionState` /
 `InitializationMode` enums.  The Pipeline owns three Representations,
@@ -214,7 +214,7 @@ LiverResections/
 
 Migrate the three Markups-derived nodes' *display* path to LayerDM
 per [ADR-0002](0002-migrate-to-slicerlayerdm.md) and
-[ADR-0013](0013-layerdm-pipeline-pattern.md), but leave the data
+ADR-0013, but leave the data
 nodes inheriting from `vtkMRMLMarkupsNode` and the interaction layer
 on `vtkSlicerMarkupsWidget`.
 
@@ -241,7 +241,7 @@ target liver mesh that seeds the same Bezier fit.  Keeping them in
 LiverMarkups would preserve a module that exists only to host two
 data carriers whose entire lifecycle is governed by the resection
 workflow — re-entrenching the implicit-state-via-scene-contents
-pattern [ADR-0013](0013-layerdm-pipeline-pattern.md) §4 retires.
+pattern ADR-0013 §4 retires.
 
 ### C. Three separate modules for the three primitives
 
@@ -274,7 +274,7 @@ the workflow they share.
   workflow.  T2 reviewers and `/slicer-review` evaluate the
   interactive surface against a single yardstick instead of three
   separate Markups subclasses.
-- **First concrete instantiation of [ADR-0013](0013-layerdm-pipeline-pattern.md)
+- **First concrete instantiation of ADR-0013
   and [ADR-0011](0011-sct-terminology-dispatch.md).**  The Pipeline
   consumes SCT triples for colour and label decoration (replacing the
   hardcoded `HepaticContourColor` / `PortalContourColor` constants in
@@ -337,7 +337,7 @@ The T2 LiverResections (all-in) phase lands in this order:
    `Set/Get` accessors, `WriteXML`/`ReadXMLAttributes` plumbing,
    and `Modified()` events on transition.
 4. **LiverBezierSurfacePipeline + three Representations** per
-   [ADR-0013](0013-layerdm-pipeline-pattern.md) — Python, observing
+   ADR-0013 — Python, observing
    both the display node and `vtkMRMLLiverResectionNode`'s state
    machine.  Ships with the three-tier test set (module test,
    per-Representation unit tests, workflow test using the
@@ -395,7 +395,7 @@ The T2 LiverResections (all-in) phase lands in this order:
     the current `LiverMarkups` constructors.
   - [ADR-0012](0012-layerdm-migration-v2-scope.md) (amended) — the
     v2.0.0 migration map collapses to T2 + T3 in light of this ADR.
-  - [ADR-0013](0013-layerdm-pipeline-pattern.md) — the canonical
+  - ADR-0013 — the canonical
     Pipeline + Representation shape this ADR instantiates; §4 is
     the state-aware-Pipeline pattern.
   - [ADR-0015](0015-cpp-algorithm-library.md) — the C++ algorithm
--- a/Docs/adr/0014-livermarkups-dissolution.md
+++ b/Docs/adr/0014-livermarkups-dissolution.md
@@ -148,9 +148,8 @@ directly — *not* `vtkSlicerMarkupsWidget`.  Free design space:
   init-mode drop-back), free of `vtkSlicerMarkupsWidget::WidgetEvent
   Menu` / `populateContextMenu` constraints.
 
-Per-role glyph rendering wires through the four existing custom OpenGL
-mappers in `LiverMarkups/VTKWidgets/` (relocated to `LiverResections/
-VTKWidgets/` alongside the widget):
+The four existing custom OpenGL mappers in `LiverMarkups/VTKWidgets/`
+relocate to `LiverResections/VTKWidgets/` alongside the widget:
 
 - `vtkOpenGLBezierResectionPolyDataMapper` — the surface.
 - `vtkOpenGLSlicingContourPolyDataMapper`,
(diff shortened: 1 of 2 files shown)


Docs/adr/0014-livermarkups-dissolution.md
```diff
@@ -0,0 +1,415 @@
...
+Per-role glyph rendering wires through the four existing custom OpenGL
+mappers in `LiverMarkups/VTKWidgets/` (relocated to `LiverResections/
+VTKWidgets/` alongside the widget):
+
+- `vtkOpenGLBezierResectionPolyDataMapper` — the surface.
+- `vtkOpenGLSlicingContourPolyDataMapper`,
+  `vtkOpenGLDistanceContourPolyDataMapper` — the init-mode rings.
+- `vtkOpenGLResection2DPolyDataMapper` — the resectogram-flattened
+  surface (carried over into the T3 Resectogram migration; relocated
+  here only to keep the migration map clean).
```
RafaelPalomar: Let's not commit to this now. It may very well be that the glyph is the same, but a halo highlights the selected control points.

Docs/adr/0014-livermarkups-dissolution.md
```diff
@@ -0,0 +1,415 @@
...
+
+After the Init→Planning transition, the two (SlicingPlane) or
+two-or-more (DistanceSpheroid) init control points **persist** on the
+parent resection's data.  A Planning→Init drop-back recovers them
+losslessly; a save/load round-trip in either state captures the
+surgeon's intent across the state boundary.  Cost on the data node is
+negligible (a handful of `double[3]` arrays per resection); the
+alternative — discarding the init control points at the state
+transition — gives up recoverable workflow intent for no payoff.
+
```
RafaelPalomar: We should have a discussion around this. While this could be true, I don't think we should leave the user to transition from resection to initialization. Sometimes the geometry of the resection can be far away than the initialization starting the resection.

Docs/adr/0014-livermarkups-dissolution.md
```diff
@@ -0,0 +1,415 @@
...
+  interior).
+- Init-mode control points (2 for SlicingPlane, ≥2 for
+  DistanceSpheroid).
+- State-machine fields (`ResectionState`, `InitializationMode`) as
+  typed enums.
+
+This is a **D-class break** per [ADR-0007](0007-version-numbering-policy.md)
+(symmetric round-trip with v1 `.lrp.fcsv` files is no longer clean);
+it is already on the v1→v2 ticket and contributes to the (D) trigger
+ADR-0007 §"Mapping the v1 → v2 jump" enumerates.
```
RafaelPalomar: We should consider other avenues for storing the data. The fcsv was very much driven by the use of markups as a base, but now we have other options.

Docs/adr/0014-livermarkups-dissolution.md
```diff
@@ -0,0 +1,415 @@
...
+```
+LiverResections/
+  MRML/
+    vtkMRMLLiverBezierSurfaceNode.{h,cxx}        # data; 4×4 grid + ring-role + weakrefs
+    vtkMRMLLiverBezierSurfaceDisplayNode.{h,cxx} # visibility, opacity, terminology (ADR-0011)
+  LiverBezierSurfacePipeline.py                   # LayerDM Pipeline (ADR-0013)
+  Representations/
+    SlicingPlaneInitRepresentation.py
+    DistanceSpheroidInitRepresentation.py
+    BezierPlanningRepresentation.py
```
RafaelPalomar, line 254: I wonder why don't wemake these cpp files? What's the advantage of keeping these python files?

Docs/adr/0015-cpp-algorithm-library.md
```diff
@@ -0,0 +1,378 @@
...
+  the corner correspondence is well-posed.
+
+Today's implementation lives in three Python functions in
+`Liver/Liver.py`:
+
+- `fit_bezier_surface` (≈ line 1914) — top-level dispatcher; takes a
+  ring + mode and returns the 4×4 control grid.
+- `runSurfacefromCurve` (≈ line 1961) — the SlicingPlane path.
+- `runSurfacefromEFD` (≈ line 2042) — the DistanceSpheroid path, with
+  the EFD parameterisation inlined.
```
RafaelPalomar, line 49: These algorithms ended up not being very stable. We should do a thorough review on them to find corner cases and bugs. With touches also testability

RafaelPalomar (commented):
Need some work

Docs/adr/0014-livermarkups-dissolution.md
RafaelPalomar: Relaxed in `59b4494`. §3 now commits to *relocating* the four custom OpenGL mappers, **not** to any specific visual treatment for selection state. Per-role glyph shapes / halo overlays / outline accents / colour modulation are deferred to the T2 implementation per ADR-0009 §3 (design rationale). The widget's right-click / right-drag event-table semantics decided in §3 are unaffected — those are interaction commitments, not visual ones.

Docs/adr/0014-livermarkups-dissolution.md
RafaelPalomar: Reframed as OPEN in `59b4494`. §4 no longer commits to persistence. Two paths laid out for separate discussion:

- *No drop-back.* Init→Planning is one-way for v2.0.0; init control points discarded at transition; Slicer scene-state undo handles "return to before the fit" if needed.
- *Drop-back with persistence.* What the previous draft committed to.

Resolution feeds §5 schema design and the state-machine in ADR-0013 §4. Decided by separate discussion before the T2 PR lands; outcome captured by amendment to this ADR.

Docs/adr/0014-livermarkups-dissolution.md
RafaelPalomar: Reframed as OPEN in `59b4494`. The **D-class break** stays committed; the **format mechanism** does not. Three candidates listed:

- *Extended `.lrp.fcsv`* — legacy-friendly, schema bump in place.
- *New MRML storage node* (`vtkMRMLLiverBezierSurfaceStorageNode`) — purpose-built; cleaner schema; XML for geometry + sidecar for high-volume payload.
- *JSON sidecar to the scene MRML* — lightweight, inspection-friendly; closer to the Segmentations-module convention.

The `.lrp.fcsv` format was Markups-driven; LiverMarkups dissolution reopens the choice. Decided by separate discussion before T2; outcome captured by amendment to this ADR or by a follow-on ADR if the format choice is substantial enough to warrant its own.

Docs/adr/0014-livermarkups-dissolution.md
RafaelPalomar: Good question — and one already adjudicated in **ADR-0013 §1** (merged via #324):

> Each `vtkMRMLLiver<Module>DisplayNode` has exactly one Pipeline subclass of `vtkMRMLLayerDMScriptedPipeline`. Pipelines are **Python**, per ADR-0004 — they hold business logic, observe MRML events, and orchestrate VTK assemblies. LayerDM's first-class Python API makes this the natural side of the seam.

Same logic applies to Representations per ADR-0013 §6 (composable Python units, unit-testable in isolation). The rationale catalogued in ADR-0004 §2:

- Pipeline / Representation iteration is seconds in Python vs. 10–60 min Slicer rebuild in C++ — and these are exactly the layers we iterate on most.
- LayerDM ships `vtkMRMLLayerDMScriptedPipeline` precisely so this layer can be Python.
- C++ remains the right home for **data-only MRML nodes** (`vtkMRMLLiverBezierSurfaceNode`, `vtkMRMLLiverBezierSurfaceDisplayNode`) and **performance-justified algorithm libraries** (which is exactly what ADR-0015 commits the math to).

So inside `LiverResections/` the split is intentional: `.py` for Pipeline + Representations + module orchestration; `.{h,cxx}` for the data-only nodes, the Bezier widget (`vtkLiverBezierWidget`), the OpenGL mappers, and the new `Algorithm/` library.

If you want to revisit ADR-0013 §1 / ADR-0004 §2, that needs a superseding ADR rather than divergence inside this one — happy to draft if you'd like to re-open the boundary. Otherwise this ADR's `.py` file extensions just track the existing commitment.

Docs/adr/0015-cpp-algorithm-library.md
RafaelPalomar: Strengthened in `59b4494`. Added **bug-discovery** as the *first* pain in the motivation list (now four pains, not three):

> The lift is also a deliberate bug-hunt opportunity. Each lifted function gets a focused C++ unit-test surface; corner cases that the Python implementation silently mishandled (degenerate rings, near-coincident control points, EFD reconstructions that diverge at high harmonic orders) become individually reproducible. Characterisation-first per ADR-0003 pins current behaviour including its bugs; the lift then either inverts the assertion for the corrected C++ behaviour or documents the divergence — in either case the bug is now visible and triaged.

Paired with the existing testability pain so the two reinforce each other: testability *and* discoverability improve together.

The characterisation surface itself landed in **PR #330** (T1-A) earlier today, pinning `fit_bezier_surface`, `elliptic_fourier_descriptors`, `inverse_transform`, `calculate_dc_coefficients` at `rtol=1e-12, atol=1e-12`. Once that merges, every corner case the lift uncovers either keeps the assertion (Python and C++ agree) or breaks it (and we have a triaged bug).

RafaelPalomar:
Review pass 1 addressed in `59b4494`. Summary:

- **ADR-0014 §3** — selection feedback deferred to T2 implementation; this ADR commits to mapper relocation, not to glyph shapes vs halo vs outline accents.
- **ADR-0014 §4** — drop-back persistence reframed as OPEN. Two paths laid out (no drop-back / drop-back with persistence); resolution feeds §5 schema design.
- **ADR-0014 §5** — storage format reframed as OPEN. D-class break stays committed; mechanism choice between extended `.lrp.fcsv` / new MRML storage node / JSON sidecar is now explicit.
- **ADR-0014 file layout `.py` Representations** — replied inline; ADR-0013 §1 + ADR-0004 §2 already commit Pipelines + Representations to Python. Happy to draft a superseding ADR if you want to re-open that boundary.
- **ADR-0015 §Context** — bug-discovery motivation strengthened. Paired with testability as the first pain in the motivation list; specific corner-case classes called out.

Inline threads can be resolved as you re-review. The two genuinely material open questions (§4 + §5) are now flagged inside the ADR rather than buried in the PR thread — their resolutions will be tracked against this ADR before the T2 PR lands.

PR #340, opened 2026-05-16
DOC: ADR-0015 — correct Eigen acquisition path (via ITK bundle, not Slicer superbuild)

RafaelPalomar:
## Summary

One-paragraph correction to ADR-0015 §2. The ADR claimed Eigen is reached via a `Slicer_USE_Eigen` superbuild toggle; that toggle does not exist on the public Slicer surface. Eigen ships privately inside ITK as the bundled `ITKInternalEigen3` header-only tree, and Slicer extensions reach a public `Eigen3::Eigen` target by pointing `find_package(Eigen3 CONFIG)` at `\${ITK_DIR}/ITKInternalEigen3-build`.

This misclaim caused the first CI failure on PR #334 (Stack 1 algorithm library) — the agent followed the ADR wording literally and `Eigen/Core` failed to resolve. Fix landed there in commits `9f4c715` + `a640f0b`. This PR updates the ADR so the next stack's agent does not retread the same misstep.

The conclusion of §2 — *"no new external dep is introduced"* — still holds: ITK is mandatory in any Slicer extension and Eigen comes through it transitively. Only the *mechanism* is corrected.

## Test plan

- [x] No code change, no CI gate beyond docs-lint.
- [x] Reproducible against PR #334's CMakeLists (`LiverResections/Algorithm/CMakeLists.txt:39-44`) which uses the same `HINTS \${ITK_DIR}/ITKInternalEigen3-build` form documented here.

## UX impact

N/A — non-UI change.

---

*AI-assisted authorship: this PR was drafted with help from Anthropic's Claude (Opus 4.7, \`claude-opus-4-7\`) via Claude Code. Opened for human review.*

PR #327, opened 2026-05-15
CI: skip Slicer build/test on docs-only pull requests

RafaelPalomar:
## Summary
Save the ~6 min Slicer build + CTest run on PRs that touch only documentation. The `build-test (slicer-main, Linux)` job still spawns and reports `success` — only the inner build/test steps are gated.

## Why now
- Slicer build is ~6 min per PR. The last week's wave of ADR PRs (#301, #302, #303, #304, #314, #317, #318, plus the in-flight #323, #324, #326) were all pure documentation. Each ran the full Slicer build for zero verification value. The savings compound.
- This is the right time before T2 begins — once code PRs are flowing, the savings shift to the new docs-only ADR amendments that will accompany them.

## Approach
**Step-level conditionals inside the `build-test` job** rather than workflow-level `paths-ignore`. Reason: preserves the check name `build-test (slicer-main, Linux)` reporting `success`. If `preview` later gets `required_status_checks` (currently `null` — verified) the rule still passes for docs-only PRs, no reconfiguration needed.

Uses [`dorny/paths-filter@v3`](https://github.com/dorny/paths-filter) with `predicate-quantifier: every` — the filter satisfies only when *every* changed file matches the pattern (i.e. truly docs-only). A single non-docs change brings the full build back.

## Filter scope (what counts as docs-only)
- `Docs/**` — ADRs, architecture diagrams, surveys, UI workflows
- root-level `*.md` — README, CONTRIBUTING, SECURITY, NOTICE
- `.github/pull_request_template.md` and `.github/CODEOWNERS`
- `CITATION.cff`, `License.txt`, `COPYRIGHT.txt`

Notably **not** in the filter (i.e. still trigger build-test):
- `.github/workflows/**` — workflow changes can affect CI behaviour
- `Resources/**` — terminology JSONs are runtime assets
- `Testing/**`, `CMakeLists.txt`, `conftest.py`, `pytest.ini`
- any module-source tree (`Liver*/**`)

## Push events
The detect step is PR-scoped (`if: github.event_name == 'pull_request'`). Push events to `preview` (merges, direct pushes, `workflow_dispatch`) always run the full build — defensive against any drift between PR-time and merge-time state.

## Worked example
PRs #314 (PR template), #317 (workflow diagrams), #318 (a11y survey) would all have skipped. ~18 min of runner time saved across that trio alone.

## Test plan
[...]

PR #343, opened 2026-05-16
FIX: docs-only paths filter — single brace-expanded pattern (PR #327 regression)

RafaelPalomar:
## Summary

PR #327 added a docs-only skip on the `build-test (slicer-main, Linux)` job using `dorny/paths-filter@v3` with `predicate-quantifier: every` and a list of glob patterns. The intent was *"every changed file matches at least one of these patterns"* — i.e. the union. But `every` in `dorny/paths-filter@v3` actually means *"every changed file matches every pattern in the list"*, which is impossible for any single file (it can be under `Docs/` or be a root `*.md`, never both).

**Surfaced by PR #340** (the ADR-0015 Eigen amendment): one file change at `Docs/adr/0015-cpp-algorithm-library.md`; action log reported `Filter docs-only = false / Matching files: none`; the full ~7-min Slicer build ran on a docs-only PR.

Fix: consolidate the list into a single brace-expanded pattern. Under `every` this becomes *"every changed file matches this one (alternation) pattern"* — the intended semantic. Brace expansion is flattened to a single level to dodge picomatch nested-brace edge cases.

## Test plan

- [x] Workflow YAML lints OK locally.
- [ ] This PR itself touches only `.github/workflows/ci.yml` — once the fix lands on `preview`, **the next docs-only PR self-tests** that the build is skipped.
- [ ] This PR is *not* docs-only (it edits the workflow), so it correctly runs the full build. Confirms the negative case still works.

## UX impact

N/A — non-UI change.

---

*AI-assisted authorship: this PR was drafted with help from Anthropic's Claude (Opus 4.7, \`claude-opus-4-7\`) via Claude Code. Opened for human review.*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,6 +159,13 @@ jobs:
       # Filter is PR-scoped — push events (merge to preview, direct push,
       # workflow_dispatch) always run the full build to catch any drift.
       # ---------------------------------------------------------------------
+      # dorny/paths-filter@v3 with `predicate-quantifier: every` matches
+      # *every* changed file against *every* pattern in the filter list
+      # (rather than the union of patterns), so listing the docs paths as
+      # separate list entries fails for any single file that doesn't satisfy
+      # all of them simultaneously — surfaced by PR #340 (one ADR change,
+      # didn't skip).  Consolidate to a single brace-expanded pattern so the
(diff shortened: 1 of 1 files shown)


RafaelPalomar (commented):
LGTM

PR #347, opened 2026-05-16
BUG: Lint workflow needs fetch-depth: 0 for pre-commit --from-ref

RafaelPalomar:
## Summary

PR #344 added a ternary that passes `--from-ref <base> --to-ref <head>` to `pre-commit/action` on `pull_request` events, restricting hooks to PR-touched files per ADR-0016 §"Cutover discipline". But pre-commit implements `--from-ref` by shelling out to `git diff --name-only <base>..<head>` to compute the file list, and `actions/checkout`'s default shallow clone (`fetch-depth: 1`) doesn't include the base commit in the local clone — `git diff` errors out.

**Surfaced by PR #345** (ADR-0014 Bezier rename amendment) after it was force-push-rebased onto post-#344 preview: the lint run reported

```
CalledProcessError: command: ('git', 'diff', '--name-only', '--no-ext-diff', '-z',
  'eff1bdfa...<head>')
```

and exited code 3.

## Fix

Set `fetch-depth: 0` on the `actions/checkout` step in `.github/workflows/lint.yml`. Full history adds a few seconds to the clone but makes the pre-commit `--from-ref` path reliable across any base/head pair.

## Test plan

- [x] **Self-testing**: this PR's own Lint check fetches deeper history; pre-commit sees only the workflow file in the diff (clean YAML), hook chain should pass.
- [ ] **After merge**: PR #345's Lint, when re-triggered against post-this-merge preview, should pass cleanly on the single touched ADR file.

## UX impact

N/A — non-UI change.

---

*AI-assisted authorship: this PR was drafted with help from Anthropic's Claude (Opus 4.7, \`claude-opus-4-7\`) via Claude Code. Opened for human review.*

PR #341, opened 2026-05-16
ENH: T2 Stack 2 / T2.1 — vtkMRMLBezierSurfaceNode + DisplayNode (ADR-0013 §8 + ADR-0014 §1)

RafaelPalomar:
## Rename: drop `Liver` prefix from Bezier-surface classes (commit `ea68438`)

One additional commit on top of `fb3eab4` renames the two new MRML classes to drop the `Liver` project-scope prefix:

- `vtkMRMLLiverBezierSurfaceNode` → `vtkMRMLBezierSurfaceNode`
- `vtkMRMLLiverBezierSurfaceDisplayNode` → `vtkMRMLBezierSurfaceDisplayNode`

Scene tag names rename in lockstep (`"LiverBezierSurface"` → `"BezierSurface"`, `"LiverBezierSurfaceDisplay"` → `"BezierSurfaceDisplay"`). Safe to change because PR #341 hasn't merged; no scenes saved with the old tag name exist anywhere.

**Rationale.** The `Liver` prefix carried no semantic weight on these classes beyond project-scope decoration — project scope is already encoded by the file path (`LiverResections/MRML/`). `LiverResection` stays as a compound clinical noun (the project's defining concept); only the generic-geometric-primitive Bezier-surface MRML classes drop the prefix. The forthcoming storage node from §5 of ADR-0014 will be born as `vtkMRMLBezierSurfaceStorageNode` in its own PR. The other project-scope-prefixed helpers around the Bezier surface (`LiverBezierSurfacePipeline`, `vtkLiverBezierWidget`) are unaffected — they remain as named in ADR-0014.

**Paired ADR amendment.** ADR-0014 (which named the old form) is amended in **PR #345** (`feature/adr-0014-bezier-rename`).

**Files touched** (10): four MRML source files (`.h`/`.cxx` × Node + DisplayNode), two test files (`Test1.cxx` for both), two `CMakeLists.txt`, and the Python wrapper test (also renamed to `test_bezier_surface_node.py` in follow-up commit `cfecd25` to match the project's existing pytest filename convention — `test_bezier_characterization.py`, `test_scaffold_smoke.py`, etc. — all prefixless).

A local `cmake --build` of the MRML library target against `/home/rafael/src/Slicer-Liver-stack2-build/` succeeded; everything links.

---

## Amendment summary (commits since first push)

[...]

PR #333, opened 2026-05-15
DOC: ADR-0014 — resolve open questions §4 (init-points) + §5 (storage)

RafaelPalomar:
## Summary
Resolves the two `OPEN` questions ADR-0014 carried out of PR #326's review pass. Both feed the T2 LiverResections all-in implementation scoping.

## §4 — Init control-point lifecycle: **read-only persistence**

Init control points (2 for `SlicingPlane`, 2-or-more for `DistanceSpheroid`) **persist on the resection's data node** as read-only audit data:
- Save/load fidelity — round-trip recovers the geometry that originally seeded the Bezier fit
- UI annotation — the resection panel surfaces "what fit produced this geometry" (display treatment deferred to T2 per ADR-0009 §3)

**There is no Planning→Init transition.** Once Planning is reached, the Bezier 4×4 grid is the sole editable representation. To re-initialise, the surgeon creates a *new* resection.

Rationale: the Bezier grid in Planning may have moved arbitrarily far from the original init ring; a Planning→Init drop-back would either silently overwrite the surgeon's planning work or surface a misleading "this ring is what you started from" affordance. The read-only persistence captures the audit value (which previously came implicitly from the always-on Markups contour nodes) without re-introducing the drop-back hazard.

Cost: 6-12 `double[3]` per resection on the data node. Negligible.

## §5 — Storage: **new MRML storage node + JSON on-disk format**

New C++ MRML storage node `vtkMRMLLiverBezierSurfaceStorageNode` reads and writes **`.lrp.json`** files — a purpose-built schema for the new data model.

JSON chosen over XML or extended `.lrp.fcsv`:
- Inspection-friendly (any text editor)
- Slicer Segmentations-module precedent
- Explicit `"schemaVersion"` field
- Single portable file per resection preserves the historical surgeon-to-surgeon plan-sharing workflow

Rejected alternatives:
- *Extended `.lrp.fcsv`* — contorts a Markups-derived CSV format to carry structured state
- *JSON sidecar to scene MRML* — split-brain coordination cost not earned by the ~70-doubles payload

**v1→v2 migration**: storage class detects legacy `.lrp.fcsv` on load and dispatches a one-way conversion. Legacy reader load-only; legacy writes unsupported.

Schema details, the migration code path, and optional richer metadata (timestamps, surgeon ID, clinical context fields) land in the T2 implementation PR.

## What does NOT change

- §1 (init-mode struct-shaped subordinate data)
- §2 (state-aware Pipeline + three Representations)
[...]

PR #350, opened 2026-05-16
ENH: Enforce init-data read-only after Init→Planning transition (ADR-0014 §4)

RafaelPalomar:
## Summary

Enforces the ADR-0014 §4 read-only-audit-data contract on `vtkMRMLBezierSurfaceNode` at the in-memory API level — the storage class (T2.5) will independently enforce the same shape on load.

Every Init-mode subordinate-data setter (`SetSlicingPlaneInitPoint`, `SetSlicingPlaneOrigin`, `SetSlicingPlaneNormal`, `SetNumberOfDistanceSpheroidInitPoints`, `SetDistanceSpheroidInitPoint`, `SetDistanceSpheroidCenter`, `SetDistanceSpheroidRadiusX/Y/Z`) short-circuits with a `vtkWarningMacro` and no `Modified()` once `State == Planning`.  Macro-generated setters (`vtkSetVector3Macro`, `vtkSetClampMacro`) were expanded into explicit member functions so the file-local `LIVER_BEZIER_GUARD_INIT_ONLY` guard can sit at the top of each body.

`SetState` gains the matching one-way invariant: `Init → Planning` is allowed, `Planning → Init` emits a warning and short-circuits.  Same-state self-assign remains a no-op.

A private `LoadingFromXML` flag — flipped for the duration of `ReadXMLAttributes` — exempts XML deserialisation from the per-setter guards, so a scene serialised with `state="Planning"` deserialises cleanly.  The guard applies to public-API mutation; XML load is internal load.

## Spec

ADR-0014 §4 (resolved by PR #333):

> After the Init→Planning transition, the init control points (two for `SlicingPlane` mode, two-or-more for `DistanceSpheroid` mode) **persist on the parent resection's data as read-only audit data**. […] **There is no Planning→Init transition.**

## Out of scope

- Storage-side enforcement (T2.5's job).
- The legacy `vtkMRMLLiverResectionNode` — unaffected.
- `Planning → some-third-state` transitions — not committed yet.
- The display node — display fields are independent of the audit-data lifecycle.

## Notable decisions

- **`LoadingFromXML` over a `StartModify`-style state save/restore**: the latter would mutate `State` itself across the `Superclass::ReadXMLAttributes` call, which is observably wrong if any observer is hooked.  A private flag is cleaner.
- **Explicit member functions over macro expansion via VTK's own
  `vtkSetVectorMacroPlus` machinery**: VTK does not ship a guarded-setter macro, and adding one for a one-class invariant would be over-engineering.  Expanding seven setters by hand is bounded scope.
- **Existing tests reordered, not rewritten**: tests that previously called `SetState(Planning)` then mutated init data (`testCopyContent`, `testXMLRoundTrip`, the parallel Python tests) reordered to populate init data pre-transition.  None were semantically exercising post-transition mutation — that path now lives in the dedicated read-only test.

## Test coverage (ADR-0008 §2 + §3)

- `LiverResections/MRML/Testing/Cxx/vtkMRMLBezierSurfaceNodeTest1.cxx` — new `testInitDataReadOnlyAfterPlanning`:
  - sets every gated field pre-transition and asserts it lands;
  - transitions to Planning;
  - asserts every gated setter rejects (`MTime` unchanged, value unchanged, bool-returning setters return `false`);
  - asserts `Planning → Init` rejection;
  - asserts `ControlGrid` + `InitMode` remain mutable post-Planning (guard is narrowly scoped).
- `Testing/Python/unit/test_bezier_surface_node.py` — parallel `test_node_init_data_read_only_after_planning` on the Python-wrapped surface, gated behind `pytest.importorskip` on `vtkSlicerLiverResectionsModuleMRMLPython`.

## Test plan

- [ ] CI `build-test (slicer-main, Linux)` green — `vtkMRMLBezierSurfaceNodeTest1` runs the new sub-test among nine.
- [ ] `pytest Testing/Python/unit/test_bezier_surface_node.py` passes inside the Slicer-bundled Python — sixteen pytest cases including the new read-only invariant test.
- [ ] CI `Lint` (pre-commit) — clean diff, lint runs on PR-touched files only per PR #347.
- [ ] CI `check-commit-message` — `ENH:` prefix, uppercase first word.

## References

- [ADR-0014 §4](Docs/adr/0014-livermarkups-dissolution.md) — init points lifecycle, no drop-back (primary spec).
- [ADR-0013 §4](Docs/adr/0013-layerdm-pipeline-pattern.md) — state-aware Pipeline observes the state node (context).
- [ADR-0003](Docs/adr/0003-testability-invariant.md) — characterisation-first.
- [ADR-0008 §2 + §3](Docs/adr/0008-testing-strategy.md) — layered + dual-mode tests.
- PR #333 (merged) — original resolution discussion that produced the §4 wording.
- PR #341 (merged) — landed `vtkMRMLBezierSurfaceNode` (this PR's edit target).

---

*AI-assisted authorship: this PR was drafted with help from Anthropic's Claude (Opus 4.7, `claude-opus-4-7`) via Claude Code.  Opened for human review.*

### DIFF
--- a/LiverResections/MRML/Testing/Cxx/vtkMRMLBezierSurfaceNodeTest1.cxx
+++ b/LiverResections/MRML/Testing/Cxx/vtkMRMLBezierSurfaceNodeTest1.cxx
@@ -682,6 +682,14 @@ int testInitDataReadOnlyAfterPlanning()
     }                                                                                                                            \
   } while (0)
 
+  // Every rejected setter below emits a deliberate vtkWarningMacro;
+  // wrap the block so the test driver's WITH_VTK_ERROR_OUTPUT_CHECK
+  // doesn't count those warnings as a failure.  The matching
+  // ``..._END()`` re-arms the warning-as-failure counter and asserts
+  // at least one warning fired in the block (so a silent-drop
+  // regression also surfaces as a failure here).
+  TESTING_OUTPUT_ASSERT_WARNINGS_BEGIN();
+
   // SlicingPlane mutations rejected post-Planning.
   double clobberOrigin[3] = { 100.0, 200.0, 300.0 };
   EXPECT_MTIME_UNCHANGED(node, node->SetSlicingPlaneOrigin(clobberOrigin));
@@ -704,6 +712,8 @@ int testInitDataReadOnlyAfterPlanning()
   EXPECT_MTIME_UNCHANGED(node, node->SetDistanceSpheroidRadiusX(99.0));
   EXPECT_MTIME_UNCHANGED(node, node->SetDistanceSpheroidRadiusY(99.0));
   EXPECT_MTIME_UNCHANGED(node, node->SetDistanceSpheroidRadiusZ(99.0));
+
+  TESTING_OUTPUT_ASSERT_WARNINGS_END();
 
   // Read back — every Init-mode value is still the one set pre-
   // transition.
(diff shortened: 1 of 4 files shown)


RafaelPalomar (commented):
LGTM

PR #359, opened 2026-05-16
ENH: Add DistanceSpheroidInitRepresentation (T2.2 iteration 3)

RafaelPalomar:
## Summary

Third iteration of the T2.2 stack: wires the **DistanceSpheroidInit** Representation of the LayerDM Pipeline pattern introduced by ADR-0013 §4 / ADR-0014 §2.  Active when `(state == Init, mode == DistanceSpheroid)`, this is the alternative initialisation path where the surgeon places a variable number of seed points (two or more) that imply an axis-aligned spheroid (`Center`, `RadiusX`, `RadiusY`, `RadiusZ`); the planning surface is then the intersection of that spheroid with the target liver mesh.

This iteration leaves the **SlicingPlaneInit** slot untouched — that is T2.2 iteration 2's territory and is in flight in parallel.

## What this renders

Per ADR-0014 §2 the Representation is responsible for three pieces of geometry:

1. **Variable-N control-point markers.**  Small sphere actors at each surgeon-placed init point.  The actor list grows / shrinks dynamically as `GetNumberOfDistanceSpheroidInitPoints()` changes between `update()` calls; removed actors are detached from the renderer and their strong refs are dropped.
2. **Spheroid visualisation.**  A translucent `vtkParametricEllipsoid` (`XRadius` / `YRadius` / `ZRadius`) wrapped in a `vtkParametricFunctionSource` and translated to the spheroid centre via a `vtkTransformPolyDataFilter`.  **Axis-aligned only** per `vtkLiverSpheroidRingExtractor`'s header — general-orientation rotation is deferred to a later stack iteration (the extractor's contract documents the same constraint).
3. **Ring on target mesh — DEFERRED.**  Tracked by `TODO(T2-target-mesh-weakref)`: the intersection ring needs the target liver mesh, which is not yet reachable from `vtkMRMLBezierSurfaceNode`.  Once the weakrefs to `TargetOrganModelNode` / `DistanceMapVolumeNode` / `VascularSegmentsVolumeNode` per ADR-0014 §1 land, this Representation will construct a `vtkLiverSpheroidRingExtractor`, feed it the target mesh + Center + Radii, and render the output as a ring polyline actor.

## Decoration plumbing (ADR-0014 §4)

- `GetResectionColor()` applies to spheroid + markers.
- `GetResectionOpacity()` applies to the (translucent) spheroid only.
- Markers stay at full opacity to remain visible against the translucent spheroid.

## Mapper relocation (deferred)

`TODO(T2-mapper-relocation)` markers note that the eventual swap from the generic `vtkPolyDataMapper` to `vtkOpenGLDistanceContourPolyDataMapper` (per ADR-0014 §3) is API-preserving and will land with the broader mapper relocation cohort.  No public-API change for this Representation when that swap happens.

## Pipeline wiring

`LiverBezierSurfacePipeline.initialize` populates the `REPRESENTATION_DISTANCE_SPHEROID_INIT` slot using the same dual-path import (package-relative + top-level `sys.path`) the `BezierPlanningRepresentation` uses.  Only the `TODO(T2.2 DistanceSpheroidInit)` block is touched — the `TODO(T2.2 SlicingPlaneInit)` block is left for iteration 2.

## Tests

`Testing/Python/unit/test_distance_spheroid_init_representation.py` mirrors the canonical `test_bezier_planning_representation.py` template:

- **Pure-Python layer** (works without VTK): construction, defaults, `None`-tolerance for display + data, color / opacity round-trip, idempotency memo on `(Center, Radii, init points)`.
- **VTK-mediated layer** (gated by `pytest.importorskip("vtk")`): spheroid colour / opacity, marker colour with markers-stay-opaque invariant, `vtkParametricEllipsoid` radii match the data node, marker count tracks `GetNumberOfDistanceSpheroidInitPoints()`, marker count grows from 2 to 5 + shrinks from 5 to 2 with renderer attach/detach actor counting, marker centres match init points, idempotent no-refresh on unchanged inputs.

All pure-Python tests pass locally; existing `test_liver_bezier_surface_pipeline.py` tests still pass (the populated slot does not break the previously-`None`-slot dispatch assertions).

## References

- ADR-0008 §2 §3 — unit-layer Representation testing discipline.
- ADR-0013 §4 §6 §7 — Pipeline pattern, Representations as composable VTK pipelines, file layout.
- ADR-0014 §1 §2 §3 §4 — data-node accessors, the three Representations, mapper relocation, read-only-after-Planning rules.

## Test plan

- [ ] CI green on the pytest scaffold (both offscreen + render-interactive).
- [ ] CI green on the build-test (no regression in `test_liver_bezier_surface_pipeline.py` from the populated slot).
- [ ] Lint / pre-commit clean on CI.
- [ ] Manual smoke once T2.6 wires the registration: drop into an `Init`-state Bezier node with `InitMode=DistanceSpheroid` and confirm the markers + translucent spheroid render at the placed-seed locations.

*AI-assisted authorship: this PR was drafted with help from Anthropic's Claude (Opus 4.7, `claude-opus-4-7`) via Claude Code.  Opened for human review.*

### DIFF
--- a/LiverResections/LiverBezierSurfacePipeline.py
+++ b/LiverResections/LiverBezierSurfacePipeline.py
@@ -317,14 +317,22 @@ class LiverBezierSurfacePipeline(_PipelineBase):
             BezierPlanningRepresentation(renderer=self.GetRenderer())
         )
 
+        try:  # pragma: no cover - exercised once per import path
+            from .Representations.DistanceSpheroidInitRepresentation import (
+                DistanceSpheroidInitRepresentation,
+            )
+        except ImportError:
+            from Representations.DistanceSpheroidInitRepresentation import (  # type: ignore[no-redef]
+                DistanceSpheroidInitRepresentation,
+            )
+
         self._representations[REPRESENTATION_SLICING_PLANE_INIT] = (
             SlicingPlaneInitRepresentation(renderer=self.GetRenderer())
         )
 
-        # TODO(T2.2 DistanceSpheroidInit): populate slot per ADR-0014 §2.
-        #   self._representations[REPRESENTATION_DISTANCE_SPHEROID_INIT] = (
-        #       DistanceSpheroidInitRepresentation(renderer=self.GetRenderer())
-        #   )
+        self._representations[REPRESENTATION_DISTANCE_SPHEROID_INIT] = (
+            DistanceSpheroidInitRepresentation(renderer=self.GetRenderer())
+        )
 
     def update(self) -> None:
         """Dispatch the active Representation by ``(state, initMode)``.
--- a/LiverResections/Representations/DistanceSpheroidInitRepresentation.py
+++ b/LiverResections/Representations/DistanceSpheroidInitRepresentation.py
@@ -1,0 +1,557 @@
+"""Representation active in ``(ResectionState=Init, InitMode=DistanceSpheroid)``.
+
+Renders the geometry that supports the **DistanceSpheroid** init mode
+(per ADR-0014 §2) — the alternative initialisation path where the
+surgeon places a variable number of seed points (two or more, per the
+data node's contract) that imply an axis-aligned spheroid (``Center``,
+``RadiusX``, ``RadiusY``, ``RadiusZ``); the planning surface is then
+the intersection of that spheroid with the target liver mesh.
+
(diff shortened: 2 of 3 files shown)


RafaelPalomar (commented):
LGTM

Issue #380, opened 2026-05-18
Review corner colour markers on Bezier resection surface (post-locator)


RafaelPalomar:
## Summary

The Bezier resection surface fragment shader currently draws **four coloured corner markers** (red / green / blue / yellow stripes at the UV corners) ungated by any property or flag.  See `LiverResections/VTKWidgets/vtkOpenGLBezierResectionPolyDataMapper.cxx` lines 227-244 (the `corner stripe` block in the fragment shader).

These are real features (orientation cues for the surgeon), but the current presentation reads more like dev-time scaffolding than a polished UX:

- High-saturation primaries on top of the medical surface.
- No way to toggle them off.
- No legend / no labelling.

## Plan

1. **Now (v2.0.0)** — leave the markers in place.  They're functional.
2. **v2.1.0** — introduce a **2D resectogram locator** that maps to the 3D resection in real-time.  This provides spatial orientation through a separate, less visually intrusive channel.
3. **Post-locator (also v2.1.0)** — re-evaluate the corner markers.  Two viable outcomes:
   - **Remove** them entirely if the locator covers the orientation need.
   - **Refine** them: more discreet rendering (smaller, lower-saturation, optional via a display-node property, or only visible during interaction).

## Decision criteria

After the locator lands, a brief UX review with the clinical users should answer:

- Does the locator alone resolve "which edge is which" without the 3D corner cues?
- If not, would a discreet marker variant cover the gap?
- Is per-user toggling worth wiring (display-node bool property)?

## Out of scope

This issue does **not** propose removing or refining the markers today.  The v2.0.0 release ships them unchanged.  Follow-up B for the legacy Bezier mapper (refresh + bug-fixes, in flight) explicitly leaves the corner markers alone.

## References

- [ADR-0020](https://github.com/ALive-research/Slicer-Liver/blob/preview/Docs/adr/0020-gpu-tessellation.md) — v2.1 GPU-tessellation rewrite; mentions corner markers as part of the v2.0.0 fragment-shader decoration that carries forward verbatim.  This issue is the place where v2.1 reconsiders that "carry forward" decision.
- Mapper review note (internal): captures the original "looks dev-feature-ish" observation.

---

*AI-assisted: issue drafted with help from Anthropic's Claude (Opus 4.7, \`claude-opus-4-7\`) via Claude Code, surfaced during the legacy Bezier mapper code review.*

RafaelPalomar:
[...]

PR #381, opened 2026-05-18
BUG: Fix latent ref leak + UB in legacy Bezier mapper + drop dead posMarker code

RafaelPalomar:
## Summary

Follow-up B to PR #379 (T2-mapper-relocation, merged 2026-05-18).  Four
surgical fixes to `vtkOpenGLBezierResectionPolyDataMapper`, informed
by a follow-up review of the relocated mapper.

- Closes a latent ref leak in `SetDistanceMapTextureObject` and an
  uninitialised-read UB in the `vtkInternal` ctor.
- Drops two pieces of dead code: the orphaned `posMarker` sampler
  (orphaned by 2023-03-06 commit `b6dbd4a`) and the unused public
  `Set/GetDistanceMapTextureObject` API.
- Scope deliberately narrow — the class is slated for replacement
  by the v2.1 GPU-tessellation rewrite per
  [ADR-0020](https://github.com/ALive-research/Slicer-Liver/blob/preview/Docs/adr/0020-gpu-tessellation.md)
  §"Rollout plan", so this PR is correctness + dead-code only, no
  modernisation.

### 1. Ref leak in `SetDistanceMapTextureObject`

The setter did a `vtkSmartPointer` assignment AND an explicit
`object->Register(this)`.  The smart pointer already holds the
reference; the manual `Register` added a second one with no matching
`UnRegister`.  Each setter call leaked one refcount on the texture
object.

Fix: drop the `Register(this)` call; let the `vtkSmartPointer`
member own the lifetime.

### 2. UB in `vtkInternal` ctor

`GridDivisions` (`unsigned int`) and `GridThicknessFactor` (`float`)
were not initialised by the ctor's member-initialiser list.  Reading
them before the consumer's first push to `SetGridDivisions` /
`SetGridThicknessFactor` was undefined behaviour; the shader could
sample garbage on the first frame.

Fix: initialise to `GridDivisions = 20`, `GridThicknessFactor = 9.5f`.
The `9.5f` choice is intentional — close to the shader's
`10.0 - thickness` discriminator, so the grid is faint-but-present
rather than blown out.  In practice
`vtkSlicerBezierSurfaceRepresentation3D::UpdateBezierSurfaceDisplay`
overwrites both before the first frame is drawn, so the bug was
unobservable in production — but the UB is now closed.

### 3. Drop dead `posMarker` sampler

Commit `b6dbd4a` (2023-03-06, *"ENH: Generating orientation marker
style directly from shader"*) replaced the texture-based orientation
marker with the procedural corner stripes that still ship today.
Three stranded references survived:

- `"uniform sampler2D posMarker;"` in the vertex-shader Dec.
- `"vec4 marker = texture(posMarker, uvCoordsOutput);"` in the
  fragment-shader Impl, into a `marker` var that's never read.
- `IsUniformUsed("posMarker")` + `SetUniformi("posMarker", 15)`
  binding in `SetMapperShaderParameters`.

No code anywhere writes to texture unit 15.  All three are removed.
The procedural corner colour markers in the fragment shader's
`corner stripe` block stay untouched — those are the current
orientation-marker implementation (separate review tracked in issue
#380).

### 4. Drop dead `Set/GetDistanceMapTextureObject` public API

Repo-wide grep confirms zero callers of either setter or getter.
The distance map is wired through
`vtkMultiTextureObjectHelper::CreateSeq3DFromRaw(..., texSeq=0)` in
`vtkSlicerBezierSurfaceRepresentation3D::CreateAndTransferDistanceMapTexture`,
bypassing this mapper's named setter entirely.  (The only sibling
matches in grep are the unrelated `vtkOpenGLResection2DPolyDataMapper`
class which has its own implementation.)

Removed:

- The `Set/GetDistanceMapTextureObject` declarations from the header.
- The setter/getter definitions from the `.cxx`.
- The `DistanceMapTextureObject` member from `vtkInternal`.
- The now-unused `vtkTextureObject` forward declaration + include.

## Test plan

- [ ] CI green (`build-test` job).
- [ ] Visual smoke (manual): mapper still renders; corner colour
      markers still visible; distance-map trim (clip-out) still works;
      grid overlay still works with the new default-init values.
- [ ] Pre-existing `vtkMRMLBezierSurfaceDisplayNode` / Bezier surface
      tests still pass (no MRML-layer changes here).

## References

- Parent PR #379 (T2-mapper-relocation, merged 2026-05-18) — relocated
  this mapper from `LiverMarkups/` to `LiverResections/` without
  behavioural change.
- Issue #380 — corner-marker review tracker (out of scope for this PR).
- [ADR-0020](https://github.com/ALive-research/Slicer-Liver/blob/preview/Docs/adr/0020-gpu-tessellation.md)
  §"Rollout plan" — the v2.1 GPU-tessellation rewrite that replaces
  this mapper entirely.

---

*This pull request was drafted by Claude (Anthropic) — model
`claude-opus-4-7` — on behalf of the maintainer.*

### DIFF
--- a/LiverResections/VTKWidgets/vtkOpenGLBezierResectionPolyDataMapper.cxx
+++ b/LiverResections/VTKWidgets/vtkOpenGLBezierResectionPolyDataMapper.cxx
@@ -56,7 +56,6 @@
 #include <vtkRenderer.h>
 #include <vtkSmartPointer.h>
 #include <vtkShaderProgram.h>
-#include <vtkTextureObject.h>
 #include <vtkTransform.h>
 
 //------------------------------------------------------------------------------
@@ -65,7 +64,6 @@ class vtkOpenGLBezierResectionPolyDataMapper::vtkInternal
 public:
   vtkInternal(vtkOpenGLBezierResectionPolyDataMapper* parent)
     : Parent(parent)
-    , DistanceMapTextureObject(nullptr)
     , RasToIjkMatrixT(nullptr)
     , IjkToTextureMatrixT(nullptr)
     , ResectionMargin(0.0f)
@@ -77,13 +75,14 @@ public:
     , ResectionOpacity(1.0f)
     , InterpolatedMargins(false)
     , ResectionClipOut(false)
+    , GridDivisions(20)
+    , GridThicknessFactor(9.5f)
   {
     this->RasToIjkMatrixT = vtkSmartPointer<vtkMatrix4x4>::New();
     this->IjkToTextureMatrixT = vtkSmartPointer<vtkMatrix4x4>::New();
   }
 
   vtkWeakPointer<vtkOpenGLBezierResectionPolyDataMapper> Parent;
-  vtkSmartPointer<vtkTextureObject> DistanceMapTextureObject;
   vtkSmartPointer<vtkMatrix4x4> RasToIjkMatrixT;
   vtkSmartPointer<vtkMatrix4x4> IjkToTextureMatrixT;
   float ResectionMargin;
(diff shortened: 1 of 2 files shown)


RafaelPalomar (commented):
LGTM

PR #384, opened 2026-05-18
ENH: Add visual-regression test harness (Level-2 minimal qSlicerApplication + ExternalData)

RafaelPalomar:
## Summary

- Lands the golden-image regression-testing infrastructure for Slicer-Liver: a Level-2 (minimal `qSlicerApplication`, `--no-main-window`) test harness for the legacy v2.0.0 Bezier mapper.
- Per [ADR-0003](https://github.com/ALive-research/Slicer-Liver/blob/preview/Docs/adr/0003-testability-invariant.md) (testability invariant) and [ADR-0020](https://github.com/ALive-research/Slicer-Liver/blob/preview/Docs/adr/0020-gpu-tessellation.md) §"Rollout plan" §7, characterisation tests of the v2.0.0 mapper are the regression gate for the v2.1 GPU-tessellation rewrite.
- **Scope: infrastructure + scaffolding only.** Real baseline PNGs are captured by the maintainer in a follow-up interactive session post-merge (the capture flow is user-launched, not agent-orchestrated).

## What lands

* **Per-scenario Python modules** under `LiverResections/Testing/Python/scenarios/` exposing the trio `setup_scene() / setup_camera() / setup_viewport()`:
  - `BezierSurface4x4Planning`  — default fixture, `ClipOut=0` (full surface visible, grid overlay + corner markers active).
  - `BezierSurface4x4Confirmed` — same fixture, `ClipOut=1` (resected side discarded by the fragment shader).  Uses the raw `ClipOut` setter; migrates to `SetState(Confirmed)` once ADR-0019 lands.
* **`capture_baseline.py`** — interactive Qt capture driver.  `s` saves a 4-file bundle (`.png` + `.mrml` + `.camera.json` + `.viewport.json`) to `Testing/baselines-staging/`; `q` quits without saving.
* **`replay_test.py`** — CTest replay driver.  Renders the scenario offscreen, snapshots via `vtkWindowToImageFilter`, compares against the ExternalData-resolved baseline PNG with `vtkImageDifference`.  Skips with exit 0 when no baseline is captured yet (the state at PR merge time) so CI stays green until the maintainer's follow-up commit.
* **`Scripts/upload_baseline.sh`** — bash + `gh release upload` helper.  Computes SHA-512 per staged file, uploads each as a release asset on `ALive-research/ALiveResearchTestingData` (`SHA512` release, content-addressed per the SlicerTestingData mirror pattern), rotates the matching `.sha512` stubs under `Data/Baseline/`, prints commit hints.
[...]

PR #382, opened 2026-05-18
ENH: Resection state machine (Init/Planning/Confirmed)

RafaelPalomar:
## Summary

Code-side enabler for [ADR-0019](Docs/adr/0019-resection-state-machine.md):
extends `vtkMRMLBezierSurfaceNode::ResectionState` from a 2-state
enum to the 3-state automaton `Init -> Planning <-> Confirmed`.

The `Confirmed` state freezes the control polygon and engages the
parenchyma-trim shader on the surface mapper (existing
`SetResectionClipOut` entry point on
`vtkOpenGLBezierResectionPolyDataMapper`), producing the v1-style
"clean view" the surgeon sees after confirming a plan.  Round-
trippable: `Confirmed -> Planning` re-opens the control polygon for
further editing.

### What changed

**Node + storage**
- `Confirmed = 2` added to the `ResectionState` enum.
- `SetState` enforces the ADR-0019 transition matrix:
  `Init -> Planning`, `Planning -> Confirmed`, `Confirmed -> Planning`
  allowed; `Init -> Confirmed`, `Planning -> Init`, `Confirmed -> Init`
  rejected with `vtkWarningMacro`.
- Read-only audit-data guard widens to "any post-Init state" so
  init-mode setters reject in both `Planning` and `Confirmed`.
- `.lrp.json` `schemaVersion` stays at 1; only the state-enum string
  set grows.  Reader gains the ADR-0019 forward-compatible
  unknown-state fallback to `"Planning"` so future enum extensions
  degrade gracefully in older builds.

**Representation + pipeline**
- `vtkLiverBezierRepresentation`: in `Confirmed` the glyph cloud is
  empty and `Pick` returns `PickRole_None` — widget disabled per the
  ADR-0019 per-state contract.
- `LiverBezierSurfacePipeline` dispatch table grows one row:
  `(Confirmed, *) -> ConfirmedRepresentation` (new Python class
  under `LiverResections/LiverResectionsLib/Representations/`).
- `ConfirmedRepresentation` calls `SetResectionClipOut(True)` on the
  surface mapper when the mapper exposes it; the call is gated by
  `hasattr` so the pre-T2-mapper-relocation skeleton (generic
  `vtkPolyDataMapper`) survives.  **The mapper file
  `vtkOpenGLBezierResectionPolyDataMapper.{h,cxx}` is NOT touched**
  — the in-flight mapper-refresh PR is the canonical owner.

**Tests**
- `testConfirmedStateTransitions` (C++) pins the full transition
[...]

PR #391, opened 2026-05-19
STYLE: Migrate BezierSurface4x4Confirmed scenario to SetState(Confirmed) (C1 follow-up)

RafaelPalomar:
## Summary

- Closes the **C1** follow-up from [PR #384's `/slicer-review` synthesis](https://github.com/ALive-research/Slicer-Liver/pull/384#issuecomment-4477825147): the visual-test scenario `BezierSurface4x4Confirmed.py` had a TODO anchored at *"when the v2 Bezier state machine (ADR-0019) lands, replace the manual ClipOut wiring with the state-machine call"*.
- [ADR-0019](../blob/preview/Docs/adr/0019-resection-state-machine.md) + the state-machine implementation landed in PR #382 (merged); the migration now goes through `vtkMRMLBezierSurfaceNode::SetState(Confirmed)`.
- Single-file edit (~30 LOC net) under `LiverResections/Testing/Python/scenarios/BezierSurface4x4Confirmed.py`.

## What changed

- Add a `vtkMRMLBezierSurfaceNode` (v2 data-only LayerDM node) to the scene and walk it `Init -> Planning -> Confirmed` per the ADR-0019 transition matrix. `Init -> Confirmed` is forbidden directly, so the Planning step is mandatory.
- Mirror the state on the legacy `vtkMRMLLiverResectionNode` (`SetState(Completed)`) so the v1 stack's persistence + GUI bindings stay coherent during the parallel-stack period.
- **Keep** the explicit `display.SetClipOut(True)` push as a documented fallback. Reason: the LayerDM Pipeline + `ConfirmedRepresentation` chain ([ADR-0013](../blob/preview/Docs/adr/0013-layerdm-pipeline-pattern.md) §6 + [ADR-0019](../blob/preview/Docs/adr/0019-resection-state-machine.md) §"Per-state contract") that would auto-propagate the state change to the trim shader is not engaged by the plain `qMRMLThreeDWidget` the visual-regression harness (`replay_test.py`) renders through — the legacy displayable-manager binding on the markups display node still drives the visible pixels in that non-Pipeline context.
- Replace the in-line TODO + the "v2 hasn't landed" docstring paragraph with a citation block to ADR-0019 §"Per-state contract" + the legal-path transition matrix.

## Out of scope

- No baseline re-capture. If the new state-machine path surfaces visual drift versus the prior ClipOut-only path on a future re-capture, that's a separate issue.
- No changes to `BezierSurface4x4Planning.py`, no changes to `ConfirmedRepresentation`, no changes to the state machine itself.

## Test plan

- [x] `ruff check` on the changed file — passes locally.
- [x] AST parse — passes.
- [x] Scenario module remains importable in a `--no-main-window` boot (no new dependencies; same `import slicer` surface).
- [ ] CI smoke-test job + LiverResections unit tests — to verify on green CI.
- [ ] Visual-regression replay test (when the `.sha512` baseline stub lands) — out of scope for this PR; the harness currently skips on missing baseline (see `replay_test.py`'s `_has_baseline` gate).

---

**AI authorship**: This PR was drafted with the assistance of Claude Opus 4.7 (Anthropic), via Claude Code, under maintainer review.

### DIFF
--- a/LiverResections/Testing/Python/scenarios/BezierSurface4x4Confirmed.py
+++ b/LiverResections/Testing/Python/scenarios/BezierSurface4x4Confirmed.py
@@ -7,30 +7,49 @@ trim is ON (``ClipOut`` uniform 1) — only the resected side of the
 Bezier surface remains visible after the fragment shader's discard
 runs.
 
-The Confirmed-state machine on the v2 Bezier surface node
-(ADR-0019) has not landed in ``preview`` HEAD at the time of
-writing.  This scenario therefore drives the underlying knob
-directly: the ``ClipOut`` setter on ``vtkMRMLLiverResectionNode``
-propagates to the ``uResectionClipOut`` shader uniform in
-``vtkOpenGLBezierResectionPolyDataMapper``.  When the v2 state
-machine lands the scenario should migrate from the raw setter to
-``bezier_v2.SetState(vtkMRMLBezierSurfaceNode.Confirmed)`` — see the
-in-line TODO inside :func:`setup_scene` for the migration anchor.
+Per ADR-0019 (the v2 resection state machine on
+``vtkMRMLBezierSurfaceNode``), the Confirmed state is committed via
+``vtkMRMLBezierSurfaceNode::SetState(Confirmed)`` — a legal
+``Planning -> Confirmed`` transition.  The state-machine API is the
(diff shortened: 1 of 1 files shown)


RafaelPalomar (commented):
LGTM

PR #392, opened 2026-05-19
ENH: Visual-test harness self-tests + scenario matrix doc (closes #387, #388)

RafaelPalomar:
## Summary

Closes two low-priority follow-up issues from PR #384's `/slicer-review` synthesis (https://github.com/ALive-research/Slicer-Liver/pull/384#issuecomment-4477825147):

- **Closes #387 (GAP-1)** — pytest self-tests for the visual-regression harness's pure-Python helpers (`_has_baseline`, `_resolve_baseline_png`, `_save_bundle`, plus `upload_baseline.sh`'s bundle-completeness pre-flight via a new `DRY_RUN=1` env gate).  Per ADR-0008 §"observability".
- **Closes #388 (GAP-6)** — new `## v2.0.0 target baseline matrix` section in `LiverResections/Testing/README.md` listing the nine baselines that should land before declaring v2.0.0 visual-regression coverage adequate, plus the mechanical 3-step recipe for adding a scenario.

## What lands

- `LiverResections/Testing/Python/test_harness/` — new directory:
  - `test_has_baseline.py` — 4-row truth-table coverage.
  - `test_resolve_baseline_png.py` — path-composition assertion.
  - `test_save_bundle.py` — bundle-write + partial-failure atomicity.
  - `test_bundle_completeness_preflight.py` — `upload_baseline.sh` rejection paths via subprocess + `DRY_RUN=1`.
- `LiverResections/Testing/Python/CMakeLists.txt` — new `pytest_harness_tests` CTest entry, gated on `Python3_EXECUTABLE` + pytest importability (mirrors the existing `pytest_scaffold` entry's skip-on-missing-pytest behaviour).
- `LiverResections/Testing/README.md` — new scenario-matrix section.
- `LiverResections/Testing/Scripts/upload_baseline.sh` — adds the `DRY_RUN=1` env gate so the test exercises the pre-flight without hitting `gh release upload`.

## Out of scope

No harness behaviour changes; pure-additive tests + doc.  The actual baseline capture for the matrix's nine scenarios is per-scenario maintainer work via the existing capture flow.

## Test plan

- [x] Python files are AST-parseable (validated locally).
- [x] No conflict markers, no Claude attributions in commit messages.
- [ ] CI: `Lint`, `docs-lint`, `check-commit-message`, `build-test (slicer-main, Linux)`, `coverage (slicer-main, Linux)`, `Paper Draft` all green.

## References

- PR #384 — visual-test harness scaffold; this PR is its declared follow-up.
[...]

PR #395, opened 2026-05-19
ENH: Wire visual-regression capture pipeline for bezier mapper

RafaelPalomar:
## Summary

Drive the v1 bezier mapper's distance-driven shader paths in the
`BezierSurface4x4{Planning,Confirmed}` scenarios so the captured baselines
actually exercise the SUT (margin band shader, ClipOut discard, control-
polygon overlay) rather than degenerate to a flat patch render with no
shader output.

Without these fixes the capture pipeline either crashes (texture upload
against a NULL `glGetError` function pointer) or produces identical
Planning + Confirmed renders (ClipOut visually inert because `dist[1]`
was zero on every fragment of the flat patch at z=0).

### Scenario fixes (`scenarios/BezierSurface4x4Planning.py`)

- Synthesise a 4-channel signed-distance volume of the parenchyma sphere
  (RGBA float, 1mm spacing) and bind it via `SetDistanceMapVolumeNode`.
  Channel 0 drives the margin band on the bezier patch; channel 1 drives
  the ClipOut discard — both encode parenchyma signed distance so
  ClipOut discards portions of the patch outside the sphere's z=0 disc.
- Wire `SetTextureNumComps(4)` + `SetDistanceMapVolumeNode` AFTER
  scene-add (mirroring `Liver.py:onResectionDistanceMapNodeChanged`).
  Wiring before scene-add fires `Modified` before the resection observer
  is attached, the display node's `TextureNumComps` stays at 0, and the
  texture upload reads past the buffer end (segfault).
- Expand the bezier control polygon from the default 0..30 to -30..60
  (centred on the sphere's xy centre) so the patch fully encloses the
  sphere's equatorial disc — makes the bezier-plane / contour-band
  alignment visually verifiable.
- Use the public scene-observer path (create-then-add with target organ
  set) rather than reaching into the protected `AddBezierSurface` (which
  is not Python-wrapped anyway since it's in the `protected:` section).

### Capture-driver fixes (`capture_baseline.py`)

- Show the widget + force a first render BEFORE `setMRMLViewNode`, so
  VTK's GL extension loader has populated `glGetError` and friends.
  Otherwise `vtkMultiTextureObjectHelper::CreateSeq3DFromRaw` calls
  `vtkOpenGLClearErrorMacro()` against a NULL function pointer and
  segfaults.  **Workaround for a real C++ bug** in
  `vtkSlicerBezierSurfaceRepresentation3D::CreateAndTransferDistance-
  MapTexture` — filing separately.
- Replace `QApplication.exec_()` (a no-op when Slicer's main loop is
  already running) with a nested `QEventLoop` so the capture blocks
  correctly inside Slicer's existing event loop.
- Apply camera + viewport spec directly to the live VTK renderer/camera
  in `--no-main-window` mode.  The standalone `qMRMLThreeDWidget` has
  no layout-manager binding to the MRML camera node, so MRML-side
  configuration was orphaned previously.
- Install `faulthandler` against fd 2 for future C-stack capture on
  segfault (Slicer's `--python-script` stdout-redirector lacks
  `fileno()`, so passing `sys.stderr` raises).

### Confirmed scenario simplification (`BezierSurface4x4Confirmed.py`)

- Drop the redundant direct `display.SetClipOut(True)` —
  `resection.SetClipOut(True)` propagates through the logic observer
  (`vtkSlicerLiverResectionsLogic::ProcessMRMLNodesEvents`) to the
  display node.

## Test plan

- [x] Interactive capture of BezierSurface4x4Planning produces a bundle
      with margin-band shading on the bezier patch.
- [x] Interactive capture of BezierSurface4x4Confirmed produces a bundle
      visually distinct from Planning (ClipOut discards patch fragments
      outside the sphere's z=0 disc).
- [x] Both bundles include `.png` + `.mrml` + `.camera.json` +
      `.viewport.json` (Level-2 reproducibility per
      [ADR-0003 §"Decision"](Docs/adr/0003-characterisation-tests.md))
- [x] Bundles uploaded to `ALive-research/ALiveResearchTestingData`
      SHA512 release (5 unique hashes; 8 `.sha512` stubs committed in
      `355721b`).  Re-uploadable via
      `LiverResections/Testing/Scripts/upload_baseline.sh`.
- [ ] CI replay test wired with `SlicerLiver_ENABLE_VISUAL_TESTS=ON`
      gate (follow-up).
- [x] Rebased onto current preview (commit `9d0d24c`).  Conflict in
      `BezierSurface4x4Confirmed.py` resolved by taking preview's
      post-PR-#391 `SetState(Confirmed)` form — the branch's region
      was a TODO placeholder PR #391 already filled in.
- [x] Lint clean — `# noqa: E402` annotations on `capture_baseline.py`
      post-`faulthandler` imports (`1f9a183`).  Block-comment above
      explains the constraint ruff doesn't model.
- [x] Harness self-test stubs extended with `_StubRendererCollection`
      + `_StubRenderer` so the `_live_renderer` lookup chain doesn't
      AttributeError under the pytest harness (`10f5718`).  Fixes the
      build-test regression that surfaced post-rebase when PR #392's
      `test_save_bundle.py` met this PR's `_live_renderer` call.

## References

- ADR-0003 §"Decision" — characterisation tests pin behaviour before
  refactor.
- ADR-0014 — bezier surface mapper architecture (the SUT).
- ADR-0019 — resection state machine extension (Confirmed scenario
  precursor; v2 path replaces the `SetClipOut(True)` mechanic with a
  state-machine call).
- ADR-0020 §"Rollout plan" §7 — the v2.1 GPU-tessellation rewrite is
  gated on these baselines passing on the v2.0.0 mapper.

## Follow-ups

- C++ bug filing for
  `vtkSlicerBezierSurfaceRepresentation3D::CreateAndTransferDistance-
  MapTexture` (uninitialised GL function loader crash; production
  workaround would be `if (renderWindow && renderWindow->Initialized())`
  gate, or move texture upload into the render loop).
- Bundle upload script + ExternalData hook-up.
- CI replay test landing (Xvfb-based; the headless `--no-main-window`
  capture mode used here doubles as the replay mode).

---

This change was drafted with AI assistance (Claude Opus 4.7 / claude-opus-4-7).


RafaelPalomar (commented):
LGTM

PR #398, opened 2026-05-19
STYLE: Drop pre-VTK-9.0 vtkMultiTextureObjectHelper from LiverMarkups

RafaelPalomar:
## Summary

- Drops ``LiverMarkups/VTKWidgets/vtkMultiTextureObjectHelper.{h,cpp}``, a pre-VTK-9.0 ``vtkTextureObject`` subclass that exposed a caller-specified ``texSeq`` to ``CreateSeqNDFromRaw``.  Upstream VTK 9.x ships the equivalent contract through ``vtkTextureObject::Create3DFromRaw`` (driven by ``vtkTextureUnitManager``) plus ``Activate()`` / ``GetTextureUnit()``.
- ``vtkSlicerBezierSurfaceRepresentation3D`` was the only consumer; its DistanceMap and VascularSegments members now hold plain ``vtkSmartPointer<vtkTextureObject>`` and upload via ``Create3DFromRaw``.
- Per-frame render binding is **not** changed.  ``SetUniformi("distanceTexture", 0)`` hardcoding in the two OpenGL resection mappers is a pre-existing render-time gap orthogonal to this helper deletion and belongs in a follow-up cleanup.

## ADR references

- ADR-0008 §2 — Testing strategy (C++ low-level row): the new test lives under ``LiverMarkups/VTKWidgets/Testing/Cxx`` per the established ctkTest-driver pattern.
- ADR-0003 — Testability invariant: behaviour-pinning test landed before the migration in a separate commit so the red-→-green progression is visible in history.
- N/A for ADR-0002 / LayerDM: this is a tactical cleanup ahead of the wider mapper LayerDM-isation deferred to v2.1.0.

## Conformance

- ``vtkLiverMultiTextureUnitAllocationTest1`` — pins the upstream ``vtkTextureObject`` contract the migration depends on: ``Activate()`` allocates distinct units, ``GetTextureUnit()`` reports them, ``Deactivate()`` frees them on ``vtkTextureUnitManager``, and ``Create3DFromRaw + Activate + Render`` produces no GL error against an offscreen ``vtkOpenGLRenderWindow``.
- ``grep -rln 'vtkMultiTextureObjectHelper' .`` returns empty.

## Test plan

- [ ] CI ``vtkLiverMultiTextureUnitAllocationTest1`` is green on the new commit (CI image has offscreen GL).
- [ ] The test passes on the pre-migration commit too (the contract is upstream-API behavior, the test does not exercise the deleted helper).  Reviewer-runnable: ``ctest -R vtkLiverMultiTextureUnitAllocationTest1``.
- [ ] Smoke-test in Slicer: open a Bezier resection scene with a distance-map volume; the surface still renders with the distance-map sampling visible.

## UX impact

N/A — non-UI change.

<details>

### Why the helper was obsolete

``vtkMultiTextureObjectHelper`` originated when ``vtkTextureObject`` did not yet expose ``Create3DFromRaw`` and the caller had to drive ``glActiveTexture(GL_TEXTURE0 + n) + glBindTexture + glTexImage3D`` by hand.  Upstream VTK 9.x ships ``Create3DFromRaw`` and ``Create2DFromRaw`` with the same allocate-upload-deactivate lifecycle the helper implemented, plus a ``vtkTextureUnitManager`` that allocates a free unit dynamically and reports it through ``vtkTextureObject::GetTextureUnit()``.  Caller-specified ``texSeq`` becomes anti-pattern — two textures with the same hardcoded texSeq stomp each other.

### Migration scope

The strict mechanical replacement is in this PR:

- ``vtkSmartPointer<vtkMultiTextureObjectHelper>`` → ``vtkSmartPointer<vtkTextureObject>`` for the two texture members.
- ``CreateSeq3DFromRaw(..., texSeq)`` → ``Create3DFromRaw(...)`` (texSeq dropped, allocation delegated to the unit manager).
- Header include and CMakeLists entries dropped.

### Out-of-scope (deliberately)

The two OpenGL mappers (``vtkOpenGLBezierResectionPolyDataMapper``, ``vtkOpenGLResection2DPolyDataMapper``) currently hardcode their sampler uniforms as ``SetUniformi("distanceTexture", 0)`` etc. and do not call ``Activate()`` on the underlying ``vtkTextureObject`` in their render paths.  Today's rendering only works because the texture creation step transiently binds the texture to unit 0; that is fragile.  The proper fix is to thread the ``vtkTextureObject`` references from the representation into the mappers, call ``Activate()`` per frame, and set the uniform to ``tex->GetTextureUnit()``.  Done in a follow-up so that this PR remains a clean mechanical lift; the test added here will cover the new render path when that follow-up lands.

### Test strategy

Per ADR-0008 §2 the test is a C++ low-level ctkTest-driver test (no Slicer launch, no Qt).  It uses an offscreen ``vtkOpenGLRenderWindow``; if the CI image lacks a working GL driver the test skips with a ``SKIP:`` line and exit 0.  ``SLICERLIVER_REQUIRE_OPENGL=1`` turns the skip into a hard failure for stricter pipelines.

</details>


RafaelPalomar:
Closing: this PR cannot be rebased — every path it touches is gone.

All 7 files live under `LiverMarkups/`, and that module was removed wholesale in #585 (ADR-0014 step 8). There is nothing left to apply the diff to.

**`vtkMultiTextureObjectHelper` itself survived** the module removal and now lives at `LiverResections/VTKWidgets/` (relocated by the same lift-and-shift as the mappers — see the note at `CMakeLists.txt:30-40`). It is **still in use**: `LiverResectionsLib/Representations/FlattenedSurfaceRepresentation.py:957` calls it through the `_import_texture_object_helper` seam for the resectogram distance-map upload (`CreateSeq3DFromRaw`), and `Testing/Python/unit/test_flattened_surface_representation.py` covers that path.

Since #620 the C++ mapper route (`SetDistanceMapImageData` → `BuildBufferObjects`) is preferred, so the helper path is now the fallback for mappers lacking that setter — reachable and tested, not dead.

So this PR's premise (drop the helper) needs re-argued against the current code before anyone acts on it, not simply re-filed. If the VTK-9-and-later case for removing it still holds, it is a fresh proposal that has to account for the Python seam and its test, and for what replaces the fallback.

> **Correction.** An earlier version of this comment asserted the helper was dead code with no consumers. That was wrong — the check behind it grepped only `.cxx`/`.cpp`/`.h` and missed the Python consumer above. The closure of this PR stands on the `LiverMarkups/` paths being gone; the dead-code claim does not, and no follow-up cleanup issue should be opened on that basis.

---
_Triaged with assistance from Claude (Anthropic Claude Opus 5)._

Issue #408, opened 2026-05-21
T5.2-b: Rename LiverSegments/ to VascularTerritories/ + content refactor


RafaelPalomar:
Rename the existing `LiverSegments/` module to `VascularTerritories/` per the v2.0.0 user-facing leap; refactor its existing Couinaud-territory logic to fit the Stage 3 tab structure committed in [ADR-0023](../blob/preview/Docs/adr/0023-unified-gui-stage-workflow.md).

## Scope

- Directory rename `LiverSegments/` → `VascularTerritories/`.
- Class rename `vtkLiverSegmentsLogic` → `vtkVascularTerritoriesLogic` (or equivalent — follow the no-Liver-prefix convention established by PR #341/#345).
- Refactor existing centerline-based partition logic into the Stage 3 Manual tab's hierarchical-table + master-detail-endpoints UX (per ADR-0023's Stage 3 Manual section).
- Add the Auto Couinaud tab (depends on T5.2-c LiverSegmentation/ for the TotalSegmentator wrapper; can stub initially).
- Wire the new node classes from #407.

## Reference docs

- [ADR-0023](../blob/preview/Docs/adr/0023-unified-gui-stage-workflow.md) — Decision §Stage 3
- [Docs/architecture/gui-stage-flow.md](../blob/preview/Docs/architecture/gui-stage-flow.md) — module ownership table

## Acceptance

- Module loads in Slicer under the new name.
- Existing Couinaud-style partition continues to work through the Manual tab.
- Auto tab present (stub OK pending T5.2-c).
- ExtensionsIndex / CMakeLists.txt updated for the rename.
- Migration notes for users in `Docs/migrations/v1-to-v2.md` (or wherever the v2 migration guide lands).

Part of v2.0.0 tracker #305 (T5.2 expansion). Depends on #407.

RafaelPalomar:
Landed on `preview`. Directory rename via PR #425 (git mv), symbol/content rename completed in **455a285** (`STYLE: Complete the LiverSegments -> VascularTerritories rename`). `LiverSegments/` is gone; `VascularTerritories/` is a full module (Logic `vtkSlicerVascularTerritoriesLogic`, `VascularTerritories.py`, MRML nodes, tests) wired into the top-level CMakeLists. Closing as complete.

[...]

Issue #407, opened 2026-05-21
T5.2-a: Implement vtkMRMLAbstractTerritoriesNode + StdCouinaud + Custom subclasses


RafaelPalomar:
Implement the territories class hierarchy committed in [ADR-0023 §"Class abstraction for territories"](../blob/preview/Docs/adr/0023-unified-gui-stage-workflow.md).

## Scope

- New C++ MRML node classes under `VascularTerritories/MRML/` (per the post-rename module):
  - `vtkMRMLAbstractTerritoriesNode` — abstract base with the polymorphic interface (`GetSegments`, `GetSegmentColor`, `GetLabelMap`, `GetSegmentationNode`, `GetMethod`, `GetSCTCode`).
  - `vtkMRMLStdCouinaudTerritoriesNode` — concrete subclass for the Auto path. Holds source image ref, AI backend identifier, subdivision enum, computed-at timestamp, output labelmap.
  - `vtkMRMLCustomTerritoriesNode` — concrete subclass for the Manual path. Holds centerline + endpoint node refs, groupings, segment names, output labelmap.
- `RegisterNodeClass` calls in `qSlicerVascularTerritoriesModule::setup()` (or equivalent post-rename).
- Subject Hierarchy folder placement at node-creation time (under "Vascular Territories", per ADR-0023's Subject Hierarchy section).

## Reference docs

- [ADR-0023](../blob/preview/Docs/adr/0023-unified-gui-stage-workflow.md) — Decision §"Class abstraction for territories"
- [Docs/architecture/territories-class-hierarchy.md](../blob/preview/Docs/architecture/territories-class-hierarchy.md) — UML + polymorphic interface table
- [ADR-0004](../blob/preview/Docs/adr/0004-python-cpp-boundary.md) — data nodes are C++

## Acceptance

- All three node classes compile and register with the scene.
- ctkTest unit tests cover polymorphic dispatch (Stage 4 overlay binding consumes the base class regardless of subtype).
- Python wrapper test confirms VTK wrapping works for the base + both subtypes.
- `vtkMRMLLiverVolumetryNode` (the Stage 5 partition node) is **not** part of this hierarchy — Stage 5's data model is separate per ADR-0023.

Part of v2.0.0 tracker #305 (T5.2 expansion).

RafaelPalomar:
[...]

PR #425, opened 2026-05-22
ENH: Test scaffolding for vtkMRMLAbstractTerritoriesNode hierarchy (#407, partial #408)

RafaelPalomar:
## Summary

Test-first scaffolding for the v2.0.0 vascular-territories class
hierarchy per ADR-0023 §"Class abstraction for territories" and
`Docs/architecture/territories-class-hierarchy.md`. Bundles the
directory rename from issue #408 (partial) and the C++ stub + test
files from issue #407.

Two commits, in test-first order per ADR-0027:

1. **`STYLE: Rename LiverSegments/ to VascularTerritories/`** — pure
   directory rename via `git mv` + a top-level `add_subdirectory`
   reference fix. Inner Python module name + `vtkLiverSegmentsLogic.*`
   are left as-is in this commit; the symbol rename is the follow-up
   implementer step.

2. **`ENH: Add vtkMRMLAbstractTerritoriesNode + StdCouinaud + Custom
   test scaffolding`** — three C++ MRML node stubs +
   `vtkMRMLAbstractTerritoriesNodeTest1.cxx` (9 named subtests
   pinning invariants 1–6 + 8–10) + `TerritoriesNodeWrapperTest.py`
   (4 Python-wrapper invariants including invariant 7 Subject
   Hierarchy folder placement). Test commit predates the
   implementation commit per ADR-0027.

The stub `.cxx` bodies return sentinel values so that:

- the MRML library compiles + links cleanly;
- the test executable builds + runs;
- the invariant tests fail red against the not-yet-implemented
  behaviour (SCT table, Subdivision-driven segment count, XML
  round-trip, Groupings round-trip, Subject Hierarchy placement);
- the follow-up implementer commit's red→green is verifiable in the
  PR diff.

## Invariants pinned

| # | Invariant | Test location |
|---|-----------|---------------|
| 1 | Abstract base is non-instantiable | C++ (intrinsic — no `New()`) |
| 2 | Both concrete subclasses instantiate | C++ + Python |
| 3 | Polymorphic `GetMethod()` via base pointer | C++ + Python |
| 4 | Std-Couinaud SCT codes (full 10-code table) | C++ |
| 5 | Custom `GetSCTCode()` defaults empty | C++ |
| 6 | `RegisterNodeClass` + polymorphic `GetNodesByClass` filter | C++ |
| 7 | Subject Hierarchy folder "Vascular Territories" placement | Python |
| 8 | MRML XML round-trip of `Method` + key state | C++ |
| 9 | `Subdivision` enum drives `GetNumberOfSegments()` (8/10) | C++ |
| 10 | Custom `Groupings` map XML round-trip | C++ |

## Follow-up — implementer commit

A separate commit (or sibling PR) by `liver-implementer` must:

- Fill `.cxx` bodies in the three new MRML classes (SCT-code table,
  XML attribute emission + parsing, Subdivision-driven segment count,
  CopyContent, segment / colour / labelmap / segmentation-node
  accessors).
- Add a module Logic class with `RegisterNodes()` + a
  `createStandardCouinaud()` / `createCustomTerritories()` factory
  that wires Subject Hierarchy folder placement under "Vascular
  Territories" (ADR-0023 §"MRML scene organisation").
- Optionally bundle the parent Python module + class rename
  (`LiverSegments` → `VascularTerritories`) in the same PR.

## Test plan

- [ ] CI green on the rename commit (no behaviour change — only paths).
- [ ] CI green on the test scaffolding commit (the new test target
      compiles + links; pre-existing tests untouched).
- [ ] The new `vtkMRMLAbstractTerritoriesNodeTest1` ctest registers
      and fails red on the not-yet-implemented invariants.
- [ ] `TerritoriesNodeWrapperTest.py` registers and reports failure
      under the implementer's not-yet-supplied module Logic / SH
      folder.
- [ ] Implementer commit lands; both test files turn green; full
      ADR-0023 §"Class abstraction for territories" / architecture-doc
      conformance reviewable in the PR diff.

## Open questions for the maintainer

- **Symbol rename scope on the parent module**: this PR keeps
  `MODULE_NAME = LiverSegments` inside `VascularTerritories/CMakeLists.txt`
  + the existing `LiverSegments.py` / `vtkLiverSegmentsLogic.*`
  filenames. The cleanest follow-up is a single implementer commit
  that renames the Python class, the `vtkLiverSegments*` files, and
  updates the `Liver/Liver.py` import — bundled in this PR or a
  sibling. Confirm scope before `liver-implementer` proceeds.
- **MRML kit library name** is decoupled
  (`vtkSlicerVascularTerritoriesModuleMRML`) — explicit, not derived
  from `MODULE_NAME`. Confirm this is the right contract for the
  scripted-module case.
- **Invariant 1 (non-instantiable abstract base)** is currently
  enforced indirectly (no `vtkStandardNewMacro` in the
  `vtkMRMLAbstractTerritoriesNode.cxx`); a direct
  `vtkNew<vtkMRMLAbstractTerritoriesNode>` line would fail to link.
  If a runtime "cannot instantiate abstract base" enforcement is
  preferred (e.g., explicit deletion in `New()`), the implementer
  step can swap in a sentinel-returning `New()` that emits an error.

---

*AI-assisted authorship: drafted by `liver-test-designer` (Anthropic Claude Opus 4.7, `claude-opus-4-7`) via Claude Code under the maintainer's brief.*


VascularTerritories/MRML/Testing/Cxx/vtkMRMLAbstractTerritoriesNodeTest1.cxx
RafaelPalomar, line 1: I don't like the license preamble to be merged with documentation. Two separate blocks.

VascularTerritories/MRML/Testing/Cxx/vtkMRMLAbstractTerritoriesNodeTest1.cxx
```diff
@@ -0,0 +1,413 @@
...
+  terminology dispatch, ADR-0004 Python/C++ boundary).  Lands per
+  ADR-0027 (test commit predates implementation).
+
+  Each test pins one architectural invariant.  Test cases that exercise
+  not-yet-implemented behaviour fail red against the stub .cxx -
+  intentional; the follow-up implementer commit flips them green.
+
+  Invariants pinned:
+
+    1. Abstract base non-instantiable via a runtime sentinel New() that
```
RafaelPalomar: We need to review this decision. What other abstract node implement? this pattern?

VascularTerritories/MRML/Testing/Cxx/vtkMRMLAbstractTerritoriesNodeTest1.cxx
```diff
@@ -0,0 +1,413 @@
...
+{
+  vtkNew<vtkMRMLStdCouinaudTerritoriesNode> couinaud;
+  CHECK_NOT_NULL(couinaud.GetPointer());
+  CHECK_STRING(couinaud->GetNodeTagName(), "StdCouinaudTerritories");
+
+  vtkNew<vtkMRMLCustomTerritoriesNode> custom;
+  CHECK_NOT_NULL(custom.GetPointer());
+  CHECK_STRING(custom->GetNodeTagName(), "CustomTerritories");
+  return EXIT_SUCCESS;
+}
```
RafaelPalomar, line 153: I wonder if the standard test VTK/Slicer macros don't exercise these tests already?

VascularTerritories/MRML/Testing/Python/TerritoriesNodeWrapperTest.py
```diff
@@ -0,0 +1,148 @@
+# Copyright (c) 2026, The Intervention Centre, Oslo University Hospital. All rights reserved.
+# Distributed under the OSI-approved BSD 3-Clause License.
```
RafaelPalomar, line 2: Why the preamble is different (short)?

VascularTerritories/MRML/Testing/Python/TerritoriesNodeWrapperTest.py
```diff
@@ -0,0 +1,148 @@
...
+#
+# Python-wrapper test for the v2.0.0 vascular-territories node
```
RafaelPalomar: Not sure we should go as far as to specify versions in documentation. It would make difficult to maintain

VascularTerritories/MRML/vtkMRMLAbstractTerritoriesNode.cxx
```diff
@@ -0,0 +1,44 @@
...
+// but it returns ``nullptr`` to prevent callers from accidentally
+// instantiating a partially-initialised territories node.  Concrete
+// subclasses ``vtkMRMLStdCouinaudTerritoriesNode`` and
+// ``vtkMRMLCustomTerritoriesNode`` supply real ``New()`` bodies via
+// ``vtkStandardNewMacro``.
+//------------------------------------------------------------------------------
+vtkMRMLAbstractTerritoriesNode* vtkMRMLAbstractTerritoriesNode::New()
+{
+  return nullptr;
+}
```
RafaelPalomar: We commented on this pattern earlier. Maybe this is not needed? How does Slicer MRML abstract nodes do?

VascularTerritories/MRML/vtkMRMLAbstractTerritoriesNode.h
```diff
@@ -0,0 +1,174 @@
...
+  HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+  LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+  THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+  This file was originally developed for the Slicer-Liver extension as
+  part of the v2.0.0 unified surgeon workflow (ADR-0023 §"Class
```
RafaelPalomar: no need to mention versions. Doc comes in a differetnt block

VascularTerritories/Resources/SlicerLiverColorMap.ctbl
RafaelPalomar, line 1: Is this file still actual?

VascularTerritories/MRML/Testing/Cxx/vtkMRMLAbstractTerritoriesNodeTest1.cxx
RafaelPalomar: Acknowledged — license preamble + descriptive docstring should be two separate `/* */` blocks. Will address in a style-sweep commit on this PR alongside the equivalent split in the .h/.cxx pair and the Python sibling (see comment on `TerritoriesNodeWrapperTest.py:2`).

VascularTerritories/MRML/Testing/Cxx/vtkMRMLAbstractTerritoriesNodeTest1.cxx
RafaelPalomar: Addressed — switched to `vtkAbstractTypeMacro` per Slicer-core precedent (`vtkMRMLAbstractViewNode`, `vtkMRMLAbstractLayoutNode`) in 1282671. The runtime-sentinel `New() → nullptr` is gone; compile-time non-instantiability is now enforced unconditionally and the Python wrapper still exports the class as a usable abstract base reachable via `IsTypeOf` / `SafeDownCast`.

The Test1 docstring at this line has also been updated (next commit) to point at `vtkAbstractTypeMacro` instead of the sentinel rationale.

VascularTerritories/MRML/Testing/Cxx/vtkMRMLAbstractTerritoriesNodeTest1.cxx
RafaelPalomar: Partial overlap, but I'd argue the explicit per-invariant sub-tests are still load-bearing:

- `EXERCISE_ALL_BASIC_MRML_METHODS` covers the generic MRML node API (`PrintSelf`, `Copy`, `CreateNodeInstance`, scene attach/detach, reference machinery). It is invoked on line 398 of this test against a concrete `vtkMRMLStdCouinaudTerritoriesNode` instance and gives us the baseline coverage.
- The 10 sub-tests in this file pin the *territories-specific* invariants from the ADR-0023 / architecture-doc contract: SCT-code ordering by `Subdivision` (4, 9), polymorphic `GetMethod()` dispatch (3), opt-in SCT-code semantics on the Custom subtype (5), `RegisterNodeClass` polymorphic filter (6), Subdivision-driven `GetNumberOfSegments` (9), and Groupings XML round-trip on the Custom subtype (10). None of these are exercised by the standard macros — they are class-contract invariants the macros cannot know about.

So: keep both. The macro covers MRML-API conformance; the sub-tests cover architecture conformance.

VascularTerritories/MRML/Testing/Python/TerritoriesNodeWrapperTest.py
RafaelPalomar: You're right — the short single-block preamble here is inconsistent with the long license + separate-doc-block pattern the other test files now use. Will normalise to two blocks (license header `# ... Distributed under ...` first, then a `# ----` separator + a descriptive docstring block) in the same style-sweep commit that addresses the C++ sibling at `Test1.cxx:1`.

VascularTerritories/MRML/Testing/Python/TerritoriesNodeWrapperTest.py
RafaelPalomar: Agreed — dropped the `v2.0.0` qualifier from the docstring (next commit). The ADR/architecture-doc references at the top of the file already anchor the contract by name, which is more durable than a version number that will drift the moment we tag the release.

VascularTerritories/MRML/vtkMRMLAbstractTerritoriesNode.cxx
RafaelPalomar: Addressed in 1282671 — the abstract base now uses `vtkAbstractTypeMacro` per the Slicer-core abstract-MRML-node convention (`vtkMRMLAbstractViewNode`, `vtkMRMLAbstractLayoutNode`). The `New() → nullptr` override is removed entirely; the ctor/dtor stay `protected`. Compile-time non-instantiability is now enforced by the macro instead of being a runtime curiosity.

This is the canonical Slicer-core pattern — the runtime sentinel was a hold-over from the earlier design draft when we were worried about Python wrapping; turns out `vtkAbstractTypeMacro` handles that correctly.

VascularTerritories/MRML/vtkMRMLAbstractTerritoriesNode.h
RafaelPalomar: Acknowledged — two intertwined issues here:

1. **Version reference** (`v2.0.0`): will drop per the same reasoning as the Python file (`TerritoriesNodeWrapperTest.py:4`) — the ADR-0023 anchor is more durable than a tag.
2. **License vs descriptive doc in one block**: same point as `Test1.cxx:1` — should be split into two `/* */` blocks.

Both folded into the upcoming style-sweep commit on this PR. Will normalise the .h/.cxx pair and the two test files (C++ + Python) together so the project gets a single consistent license-block convention.

VascularTerritories/Resources/SlicerLiverColorMap.ctbl
RafaelPalomar: Yes, still actual. The file is referenced in three places:

- `VascularTerritories/CMakeLists.txt:18` — packaged into the module resources.
- `VascularTerritories/VascularTerritories.py:368` — runtime path the module loads in `setup()` to seed the colour table used by the Stage-4 territories overlay.
- `VascularTerritories/MRML/vtkMRMLStdCouinaudTerritoriesNode.cxx:67` — comment naming it as the canonical visual source for the default palette so the in-class fallback stays in sync with what the surgeon actually sees.

ADR-0023 §"UI Stage 4 — Volumetry" + `Docs/architecture/ui-stage-4-volumetry.md` rely on this overlay being seeded from a deterministic palette; the .ctbl is that palette. The path was updated in c08ae01 when the file moved from `LiverSegments/Resources/` to `VascularTerritories/Resources/` as part of the module rename.

PR #441, opened 2026-06-08
ENH: Relocate distance maps + delete orphaned LiverLogic dead duplicate

RafaelPalomar:
## Summary

Completes the orphaned-domain-code relocation. The signed-Maurer
distance-maps group from the orphaned `LiverLogic` class relocates to a
dedicated `LiverResectionsLib.DistanceMaps` module owned by the
resection stage; everything else parked on that class (Bezier / EFD /
PCA / mesh-preprocessing, ~1000 lines) is a dead duplicate of the C++
`LiverResections/Algorithm/` library and is deleted, collapsing the now
import-orphaned `LiverLib` package entirely. The dead Python arm of the
Bezier characterisation test is removed; the surviving C++ arm keeps the
same `EXPECTED_*` pins. Honours ADR-0023 "Shell composition": the Liver
shell composes, it does not compute. Net: +155 / -1568 lines. Risk
surface is low — the deleted code had no live caller (verified by grep);
the one piece that genuinely relocates is pinned by a new module-layer
test.

## ADR references

- ADR-0023: Unified GUI stage workflow — realises §"Shell composition
  (Option H)" (no domain logic in the shell) by relocating/deleting the
  orphaned compute.
- ADR-0015: C++ algorithm-library lift — the deleted Bezier/EFD/PCA
  Python code is a dead duplicate of the C++ library this ADR
  introduced.
- ADR-0008 §2: Testing strategy — the new distance-maps test is
  Module-layer (MRML scene + pipeline, no Qt widgets).

## Conformance

- `LiverResections/Testing/Python/test_distance_maps_relocation.py` (3
  tests) pins the relocated compute: output-geometry preservation at
  unit down-sampling rate, signed-distance sign + spacing-weighted
  magnitude, and `imageResample` physical-extent preservation. Green
  under a launched Slicer.
- `Testing/Python/unit/test_bezier_characterization.py` C++ arm (8
  tests) re-asserts the same `EXPECTED_*` constants against the C++
  Algorithm wrapper, so deleting the Python duplicate drops no coverage.
  Green under a launched Slicer with the Algorithm module on path.
- `Liver/Testing/Python/test_liver_shell_no_domain_imports.py` (static
  AST) stays green: `Liver/Liver.py` carries no forbidden
  numerical/domain imports or algorithm calls.

## Test plan

- [x] Distance-maps relocation test green under the local Slicer (3
      passed).
- [x] Bezier characterisation C++ arm green under the local Slicer (8
      passed).
- [x] Shell no-domain-imports invariant green (AST check).
- [x] `ruff check`, `pyupgrade`, trailing-whitespace, end-of-file,
      mixed-line-ending, and the copyright-year hook pass on the changed
      files. (Node-based `prettier`/`ruff` pre-commit env install is
      unavailable offline in this environment; ran `ruff` directly and
      the project-local hook scripts as the documented fallback.)
- [ ] CI build + full ctest matrix (safety net).

## UX impact

N/A — non-UI change

<details>

The planner established that of the groups parked on the orphaned
`LiverLogic` class, exactly one genuinely relocates rather than being a
dead duplicate: the signed-Maurer distance-maps group
(`computeDistanceMaps` + `imageResample`). The Bezier surface fitting,
elliptic Fourier descriptors, PCA, and mesh-preprocessing groups all
have bit-equivalent C++ implementations in `LiverResections/Algorithm/`
(introduced per ADR-0015) and no live Python caller.

The relocation is a byte-faithful move: `computeDistanceMaps` and
`imageResample` are de-self'd to module-level functions (the only use of
`self` was `computeDistanceMaps` chaining `imageResample` on the
down-sampling path, preserved as a plain function call). The
characterisation discipline (ADR-0003) means the test pins the existing
behaviour rather than a refined version, so the body is intentionally
left verbatim.

Deletion safety was confirmed by grepping the whole repo for `LiverLogic`
and `legacy_logic`: the only surviving references are documentation
comments and the C++ Algorithm provenance annotations (which describe
where the C++ lift came from). No runtime code imports `LiverLib` or
constructs `LiverLogic`, so collapsing the package is safe.

The Bezier test's `_bernstein_degree_3` / `_make_bezier_fixture` /
`_make_contour_fixture` helpers were retained even though the planner's
annotation described them as "now-unused": the surviving C++ arm shares
all three, so they are not Python-arm-only. This was the one deviation
from the plan's literal wording, resolved in favour of keeping the C++
arm working.

The "v2.1 deliverable" wording for the Anatomy-Definition
(LiverSegmentation) stage was corrected to describe gating on the
module's presence in the build rather than on a future release;
LiverSegmentation is a v2.0 module. The graceful-degradation behaviour
when the module is absent is unchanged.

</details>


LiverResections/Testing/Python/test_distance_maps.py
RafaelPalomar, line 1: Relocation (refactor) is circumstantial. I don't think it deserves a test, at least with that name. In addittion, it looks that this test is more than about relocation of the code, but does some fundamental testing. I would suggest renaming

RafaelPalomar: Renamed to address this — the relocation framing is gone:

- File: `test_distance_maps_relocation.py` → `test_distance_maps.py`
- CTest entry: `LiverResections_distance_maps_relocation` → `LiverResections_distance_maps`
- Docstrings + skip reasons reframed around the durable invariant — the distance-map **computation contract** (unit-rate output geometry preservation; signed-Maurer sign + spacing-weighted magnitude at sampled voxels; `imageResample` geometry) — rather than the circumstantial move. Dropped the now-stale test-first / red-state narrative, since the code already lives in `DistanceMaps.py`.

No assertion changes — the test logic is untouched. Done in 3049f9e.

PR #444, opened 2026-06-08
ENH: Launched-Slicer pytest harness (#439)

### LINKED ISSUE #439: Launched-Slicer pytest harness for widget-level tests (T5.2-d follow-up)
## Background

T5.2-d (PR #438) landed pytest-style tests for the Liver-shell
sidebar (`Liver/Testing/Python/test_liver_shell_*.py`) following
ADR-0008 §6 "Python-side testing".  The CMake harness invokes
bare `PythonSlicer -m pytest <file>`, which is the existing
Slicer-Liver pattern.

That harness is fine for the Python-side semantics tests
(`test_liver_shell_isstagecomplete.py` symbol-existence + AST
checks; `test_liver_shell_no_domain_imports.py` AST-only check),
but it cannot satisfy the widget-level tests
(`test_liver_shell_sidebar.py` T1/T4/T5/T6) which need:

  * `qt.QWidget` — present only when `qSlicerApplication` has
    been initialised.
  * `slicer.mrmlScene` — same.

Under bare PythonSlicer those tests skip cleanly (per the
`conftest.py` `_require_qt_widget` / `_require_mrml_scene`
helpers added in PR #438), so they don't break CI -- but they
also don't *run*, which means the widget-level invariants in
T5.2-d are characterised but not actively guarded.

## Out of scope for T5.2-d, in scope here

Add a launched-Slicer pytest harness that runs the widget-level
tests inside a real `qSlicerApplication`:

```
Slicer --no-splash --python-script $(which pytest) -- \
  Liver/Testing/Python/test_liver_shell_sidebar.py
```

Wiring this into CTest probably needs a new CMake macro -- the
existing scripted-module-level "run this pytest under bundled
Python" pattern does not initialise a Qt event loop.

The pattern SlicerLayerDM uses for its widget tests is a
plausible model; trame-slicer follows a similar shape.

## Acceptance

- New CTest target (e.g. `LiverShellWidgetTest`) that runs the
  T5.2-d widget-level tests in a launched Slicer with
  `qSlicerApplication` initialised.
- All four tests (T1 / T4 / T5 / T6) report PASS, not SKIP, under
  this target.
- Existing bare-PythonSlicer pattern continues to cover the
  Python-side semantics tests.

## References

- ADR-0008 §6 "Python-side testing"
- ADR-0023 §"Shell composition (Option H)"
- PR #438 (T5.2-d) — added the conftest skip helpers + the
  pattern this issue is replacing

RafaelPalomar:
## What

Closes #439. Adds a pytest path that runs **inside a launched Slicer** so tests can reach wrapped MRML bindings and a live scene — the existing `pytest_scaffold` entries run bare CPython and can't, so widget/scene tests skip.

- `Liver/Testing/Python/run_pytest_launched.py` — thin driver run via `Slicer --python-script`; calls `pytest.main()` on the project roots and routes the int return through `slicer.util.exit()`, **failing closed** on test failure, collection/import error, and `NO_TESTS_COLLECTED` (mirrors the `LiverResections/.../replay_test._exit` idiom).
- New CTest entry `pytest_launched` — launches with modules enabled (full `--additional-module-paths`, no `--disable-modules`) so transitive deps are exercised, not skipped. The bare path is preserved (additive).
- Contract pinned by `test_run_pytest_launched_contract.py` (exit-code propagation + fail-closed) and `test_bare_pytest_skip_path_preserved.py` (launched path doesn't regress bare-path skip semantics). ADR-0008 §6.

Also drops the now-obsolete `py_LiverSegmentsTest` CTest excludes (that test is `py_VascularTerritoriesTest` post-rename, with SlicerVMTK wired at the module level).

## Why

The launched harness is the missing net that would have caught the PythonQt/`mrmlScene`/leak bugs locally instead of via CI round-trips.

### DIFF
--- a/Liver/Testing/Python/test_run_pytest_launched_contract.py
+++ b/Liver/Testing/Python/test_run_pytest_launched_contract.py
@@ -120,11 +120,19 @@ def _run_driver(target_path: str):
     place to update.
     """
     driver = _require_driver()
+    # Force the driver's sys.exit path so the process exit code is
+    # faithful regardless of interpreter.  Under PythonSlicer (CI's
+    # sys.executable) ``slicer`` imports but there is no launched event
+    # loop, so ``slicer.util.exit`` cannot carry the code out; the real
+    # launched harness (pytest_launched) leaves this unset and relies on
+    # the event loop.  This self-test pins the propagation *logic*.
+    env = {**os.environ, "SLICER_PYTEST_LAUNCHED_FORCE_SYSEXIT": "1"}
     return subprocess.run(
         [_python_executable(), driver, "--", target_path],
         capture_output=True,
         text=True,
         timeout=300,
+        env=env,
     )
 
 
--- a/Liver/Testing/Python/test_run_pytest_launched_contract.py
+++ b/Liver/Testing/Python/test_run_pytest_launched_contract.py
@@ -40,14 +40,12 @@ exit code -- exactly what CTest keys off.
 Per ADR-0008 §6 (CI matrix -- "every PR runs everything", non-zero on
 any failure) and the ``replay_test._exit`` exit-propagation precedent.
 
-RED STATE
----------
-Every test in this file is expected to RED-FAIL until
-``run_pytest_launched.py`` is implemented: the driver script does not
-yet exist on disk.  The tests assert its presence and fail with a clear
-message rather than skipping, so the red state is visible (a skip would
-mask the missing driver).  The implementer (``liver-implementer``) turns
-them green by landing the driver.
+Driver presence
+---------------
+``_require_driver`` hard-fails (not skips) if ``run_pytest_launched.py``
+is ever missing, so an accidentally removed or renamed driver surfaces
+as a red test rather than a silent skip (ADR-0008 §6 -- every PR runs
+everything, non-zero on any failure).
 
 See also
 --------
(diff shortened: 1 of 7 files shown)


Liver/Testing/Python/test_run_pytest_launched_contract.py
```diff
@@ -0,0 +1,248 @@
...
+
+RED STATE
+---------
+Every test in this file is expected to RED-FAIL until
+``run_pytest_launched.py`` is implemented: the driver script does not
+yet exist on disk.  The tests assert its presence and fail with a clear
+message rather than skipping, so the red state is visible (a skip would
+mask the missing driver).  The implementer (``liver-implementer``) turns
+them green by landing the driver.
+
```
RafaelPalomar: Is this true? `run_pytest_launched` does exist

RafaelPalomar: Good catch — that was stale test-first framing. The contract test was authored before `run_pytest_launched.py` landed, so the docstring still claimed the driver "does not yet exist on disk." It exists now, so I've reframed: `_require_driver` is a defensive guard that fails loudly (not skips) if the driver is ever removed/renamed — not a test-first red marker. Same for the per-test `RED-fails until…` notes. No assertion changes.

PR #447, opened 2026-06-09
ENH: Drop blocking volumetry-complete dialog; rely on the wait cursor

### LINKED ISSUE #322: UX: replace modal `QMessageBox.information` after long ops with non-modal notification
Surfaced by the T1-F current-state survey (PR #317).

`QMessageBox.information` (modal popup) is used after long-running operations in both resection planning and volumetry. On a multi-monitor surgical workstation, the modal can land off the active panel — the user has to hunt for it before they can continue.

## Per ADR-0009 / ADR-0010
This is a user-visible behaviour change subject to the UX discipline ADR. ADR-0010 also constrains us: don't ship a parallel notification mechanism if Slicer has one — see #318 (Slicer host accessibility/UX survey) for what Slicer does.

## Options
1. **Toast notification** (transient, top-right of active panel). Qt 6 has no built-in toast; would need a small custom widget. Slicer has no toast convention per #318. Risk: parallel mechanism.
2. **Status-bar text** — already present in Slicer, low-friction, but ephemeral and easily missed.
3. **Inline panel message** — within the operation's own widget; doesn't follow the user but is always discoverable.
4. **Accept the modal** — current behaviour. Document the multi-monitor caveat.

## Decision needed
Which pattern Slicer-Liver adopts. The #318 survey result should inform the call — pick whatever aligns with Slicer host practice.

## Cross-refs
- ADR-0009 (UX discipline)
- ADR-0010 (host alignment — don't ship parallel notification mechanisms)
- #318 Slicer host survey

Tracked by #305.

---

*AI-assisted authorship: this issue was drafted with help from Anthropic's Claude (Opus 4.7, `claude-opus-4-7`) via Claude Code. Opened for human review.*

RafaelPalomar:
Closes #322.

## What
The only post-long-op `QMessageBox.information` in the codebase was `LiverVolumetry.onComputeAdvancedVolumeButtonClicked`'s *"The targeted liver volumetry was computed"* box — a **blocking modal with no actionable content**.

A wait cursor **already** wraps the computation (`setOverrideCursor(Qt.WaitCursor)` at the top of the handler, `restoreOverrideCursor()` after), so in-progress feedback is covered and the populated volumetry table is the result. This removes the redundant blocking dialog — no non-blocking notification framework needed.

## Note (out of scope, flagged)
The existing `setOverrideCursor`/`restoreOverrideCursor` pair is not exception-safe — if `computeVolume` raises, the cursor stays stuck. Pre-existing; not fixed here to keep the change minimal. Easy follow-up: wrap the compute in `try/finally`.

### DIFF
--- a/LiverVolumetry/LiverVolumetry.py
+++ b/LiverVolumetry/LiverVolumetry.py
@@ -224,8 +224,10 @@ class LiverVolumetryWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
     self.logic.computeVolume(segmentsVolumeNode, targetSegmentVolumeNode, self.ui.InputSegmentationSelector.currentNode(), outputTable, ROIMarkersList, resectionNodes)
 
+    # The wait cursor (set above) is the in-progress feedback; the
+    # populated volumetry table is the result, so no blocking completion
+    # dialog is shown (non-blocking feedback, ADR-0009 UX discipline).
     qt.QApplication.restoreOverrideCursor()
-    qt.QMessageBox.information(None, "Information", "The targeted liver volumetry was computed.")
 
     self.showTable(outputTable)
     slicer.mrmlScene.RemoveNode(segmentsVolumeNode)


RafaelPalomar (commented):
LGTM

Issue #321, opened 2026-05-15
Decide: enable `ClosedCurveButton` or remove its dead handler branches


RafaelPalomar:
Surfaced by the T1-F current-state survey (PR #317).

`ClosedCurveButton` in `ResectionsWidget.ui` has `enabled=false`, but the Python signal handler still has live branches for it. The button is unreachable in the current UI, but the code behind it isn't a no-op.

## Options
1. **Enable the button** — make it reachable. Restores the feature the handler was written for. Requires understanding whether the underlying support is intact.
2. **Remove the dead handler** — accept closed-curve resection isn't in scope; delete the handler branches and the .ui button.
3. **Document and defer** — leave both in place, document the latent capability, revisit during LayerDM migration.

## Decision needed
Likely a product call (was closed-curve resection intentionally disabled, or accidentally lost in a refactor?).

Once decided, trivial implementation either way.

Tracked by #305.

---

*AI-assisted authorship: this issue was drafted with help from Anthropic's Claude (Opus 4.7, `claude-opus-4-7`) via Claude Code. Opened for human review.*

RafaelPalomar:
Deferred to **v2.1.0** (maintainer decision 2026-06-09): the closed-curve resection initialisation method is a v2.1 feature. Enabling `ClosedCurveButton` is more than un-greying the button — it needs the init-method UI to be reconciled with the v2 Bézier init representations (`SlicingPlaneInit` / `DistanceSpheroidInit`, ADR-0014), and possibly a new closed-curve init representation. That design + wiring is v2.1 scope.

Open v2.0.0 question (separate, small): the `ClosedCurveButton` radio option is reachable in the Stage-4 Resections panel but its handler is dead — in v2.0 it should be hidden/disabled so it isn't a surgeon-facing dead-end. Tracked for a decision.

PR #448, opened 2026-06-09
ENH: Disable ClosedCurve resection radio (v2.0 dead-end)

RafaelPalomar:
## Summary

Disable the **MarkupClosedCurve** resection-initialisation radio button in
the Resections widget. The closed-curve initialisation path is not wired for
the v2.0 resection workflow — it is deferred to v2.1 — so exposing a
selectable control that silently does nothing is a UX dead-end.

This sets the radio `enabled: false` and adds a tooltip
("Closed-curve initialisation is planned for v2.1.") so the unfinished path
is visibly parked rather than removed.

## Why a separate tiny PR

ClosedCurve resection was deferred to v2.1. This is the small v2.0 loose-end
of that deferral: keep the control discoverable (so users know it is coming)
but not selectable. Re-enabling it is the entry point for the v2.1 work.

## Scope

- `Liver/Resources/UI/ResectionsWidget.ui` — one widget, `enabled` flipped to
  `false` + a `toolTip`.

Relates to #321 (does not close it — #321 tracks the full ClosedCurve
initialisation, which lands in v2.1).

---

### DIFF
--- a/Liver/Resources/UI/ResectionsWidget.ui
+++ b/Liver/Resources/UI/ResectionsWidget.ui
@@ -113,10 +113,13 @@
               <item>
                <widget class="QRadioButton" name="ClosedCurveButton">
                 <property name="enabled">
-                 <bool>true</bool>
+                 <bool>false</bool>
                 </property>
                 <property name="text">
                  <string>MarkupClosedCurve</string>
+                </property>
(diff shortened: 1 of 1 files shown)


RafaelPalomar (commented):
LGTM

PR #449, opened 2026-06-09
ENH: Stage-2 surgeon UI atop the segmentation orchestrator (LiverSegmentation)

RafaelPalomar:
## Summary for humans

Stage 2 (Anatomy Definition) gets its surgeon-facing UI, built on top of the
existing segmentation orchestrator. The panel is now a tabbed workspace: one
tab per anatomical structure (Liver parenchyma / Portal vein / Hepatic vein /
Tumors), each with the same card — Run TotalSegmentator, a status/progress
row, Accept, Reject, and Edit-in-Segment-Editor. Each tab label carries a
confirmation glyph (○ → ✓) that flips once that structure has been accepted
into the single canonical segmentation, mirroring the Liver-shell stage
indicators.

Under the hood the orchestrator grows the methods the cards drive: a `segment`
entry point that funnels every AI call through one stubbable seam, a `reject`
that discards a scratch result, input selection that always works on the
Stage-1 portal-venous volume, and an `isStructureAccepted` predicate that reads
only the canonical node. Multi-focal tumors are preserved as N separate Mass
segments rather than collapsed into one. A small Stage-2-local backend-status
row shows whether TotalSegmentator is installed and offers a pre-download
button; the AI backend is still only ever imported on the call path, never at
module load.

This revision also fixes a launched-Slicer startup failure: the module's
Python package had been placed at the module root (`__init__.py` plus
`ToolWrappers/`), which no sibling module does. When the launched-test harness
put the module source dir on `--additional-module-paths`, Slicer's
scripted-module factory tried to instantiate that root `__init__` as a module,
half-initialising the application — `mrmlScene` stayed unset, the Stage-2
scene/widget tests silently skipped, and the launched-pytest exit path hung.
The package now lives under `LiverSegmentationLib/`, mirroring every sibling
(e.g. `LiverResections/LiverResectionsLib`), so the module root holds only
scripted-module files and the factory finds no root marker. The wrapper import
collapses to one canonical name and the previous shadow-drop workaround is
removed.

This is a Draft pending review.

## ADR references

1. ADR-0024 §"Per-structure micro-workflows" — the four per-structure cards
   and the Run → scratch → Accept/Reject lifecycle.
2. ADR-0024 §"Output contract" — single canonical node; one SCT-tagged
   segment per structure; multi-focal tumors as N separate Mass segments.
3. ADR-0024 §"Lazy install for AI backends" / §Follow-on — the Stage-2-local
[...]

PR #452, opened 2026-06-09
ENH: Harden launched-Slicer exit against a drifted slicer.util

RafaelPalomar:
## Summary

Harden the launched-Slicer pytest exit path so a drifted Slicer image can
never hang CI for the full test timeout.

## The bug

`Liver/Testing/Python/run_pytest_launched.py` (and the older
`LiverResections/Testing/Python/replay_test.py`) routed pytest's exit code
through `slicer.util.exit(code)` inside a bare `except ImportError`:

```python
try:
    import slicer
    slicer.util.exit(code)
except ImportError:
    sys.exit(code)
```

A transient `slicer-main` image shipped a **partial `slicer.util` with no
`exit` attribute**. That raises `AttributeError` — *not* `ImportError` — so it
escaped the handler. The script aborted while Slicer's Qt event loop kept
running, and `pytest_launched` hung until the **1500 s CTest timeout**,
reporting the wrong status even though the underlying run was green
(`94 passed, 43 skipped in 2.66s`, then the traceback).

## The fix

Degrade across the same primitive `slicer.util.exit` itself uses
(`app.commandOptions().runPythonAndExit = False; app.exit(code)`), then
`sys.exit` as the last resort:

- `slicer.util.exit(code)` when present (unchanged happy path)
- else `slicer.app.exit(code)` after clearing `runPythonAndExit` (so Slicer
  doesn't overwrite our code) — the faithful equivalent
- else `sys.exit(code)` (standalone CPython)

A missing API now yields a coded exit, never a hang. Same one-spot fix in
`replay_test._exit`, which shared the bug.

## Conformance

Three exit-routing invariants pinned in the driver's contract self-test
(`test_run_pytest_launched_contract.py`):

| Invariant | Test |
|---|---|
| `slicer.util.exit` is preferred when present | `test_exit_uses_slicer_util_exit_when_present` |
| missing `slicer.util.exit` → `app.exit` fallback (no hang/raise) | `test_exit_falls_back_to_app_exit_when_util_exit_missing` |
| no `slicer` at all → coded `sys.exit` | `test_exit_falls_back_to_sysexit_when_slicer_unimportable` |

The fallback test reproduces the exact `AttributeError: module 'slicer.util'
has no attribute 'exit'` against the pre-fix body (verified red→green
locally).

Per ADR-0008 §6 (fail-closed CI posture) and the `replay_test._exit`
exit-propagation precedent.

---

### DIFF
--- a/Liver/Testing/Python/run_pytest_launched.py
+++ b/Liver/Testing/Python/run_pytest_launched.py
@@ -107,15 +107,41 @@ def _exit(code: int) -> None:
     ``sys.exit`` for the standalone CPython invocation (the harness
     self-test in ``test_run_pytest_launched_contract.py``, or a local
     lint run).  Same shape as ``replay_test._exit``.
+
+    ``slicer.util.exit`` is preferred, but a drifted Slicer image has
+    been observed to ship a partial ``slicer.util`` with no ``exit``
+    attribute.  Routing the code through it inside a bare ``except
+    ImportError`` let the resulting ``AttributeError`` escape: the
+    script aborted while Slicer's Qt event loop kept running, hanging
+    until the CTest timeout fired.  So degrade across the same primitive
+    ``slicer.util.exit`` itself uses -- clear ``runPythonAndExit`` so
+    Slicer does not overwrite our code, then ``app.exit`` -- and only
+    then ``sys.exit``, so a missing API yields a coded exit, never a
+    hang.
     """
     if os.environ.get(_FORCE_SYSEXIT_ENV) == "1":
         sys.exit(code)
     try:
         import slicer  # type: ignore[import-not-found]
-
-        slicer.util.exit(code)
     except ImportError:
         sys.exit(code)
+        return
+
+    util_exit = getattr(getattr(slicer, "util", None), "exit", None)
+    if callable(util_exit):
+        util_exit(code)
+        return
+
+    app = getattr(slicer, "app", None)
+    if app is not None:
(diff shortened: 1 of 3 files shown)


RafaelPalomar:
Closing without merging — the premise this PR was opened on did not hold up.

It was spawned believing the launched-harness 1500s hang came from a drifted `slicer-main` image shipping a partial `slicer.util` (no `exit`). That was disproven: the failing re-run pulled the byte-identical healthy image, and the missing `slicer.util.exit` was a **symptom** of a test poisoning `sys.modules['slicer']` in the shared launched interpreter — not an image issue.

The real fixes landed elsewhere:
- **#449** — subprocess-isolates the import-purity stub so it can't poison the shared interpreter, plus a `_no_slicer_module_leak` autouse guard that fails any future in-process leak loudly.
- **#453 (ADR-0030)** — pins the CI Slicer image, so the image can no longer drift under us, plus a `PythonSlicer` runtime smoke-check at the verify stage.

The change here (widen `_exit`'s over-narrow `except ImportError`; fall back `util.exit → app.exit → sys.exit`) remains a reasonable belt-and-suspenders — it would convert a *future* in-process-leak regression from a 25-min hang into a fast fail. But with the leak isolated (#449) and the image pinned (#453), it guards a now-remote scenario, so we're not carrying it. Keeping the branch (`feature/launched-exit-hardening`) for reference if that fail-fast hardening is ever wanted.

PR #453, opened 2026-06-09
COMP: Pin CI Slicer image to a specific build; bump deliberately (ADR-0030)

RafaelPalomar:
## Summary

Pin the CI Slicer container to a specific build instead of the rolling
`:latest` tag, and document the deliberate-bump workflow. Adds **ADR-0030**.

## Why

Both CI jobs (`build-test`, `coverage`) pulled
`ghcr.io/alive-research/slicer-build-ubuntu2404:latest` — a rolling tag the
ALive-Docker repo rebuilds to track Slicer **main**. That couples our CI to
the *instantaneous* state of an upstream dev branch.

It bit us: a `pull_request` run and a `preview` run ~50 min apart pulled
**different** `:latest` digests. The earlier one carried a Slicer-main build
whose Python runtime was broken (`slicer.util` lacked `exit`), which
manifested as a 1500 s launched-test hang on the PR while `preview` (later,
healthy `:latest`) passed. Same commit, opposite result, depending only on
*when* it ran — non-reproducible and non-bisectable.

(Note: the harness bug that *surfaced* in that incident was ours and is fixed
separately. This PR fixes the structural problem — non-reproducible CI and
uncontrolled coupling to upstream churn — not that specific bug.)

## What changed

- **Pin both jobs** to `:main-4fb4ac4` (the known-good build, digest
  `sha256:549a820f…`) — same pin for both, never `:latest`.
- **Deliberate bump workflow** (ADR-0030): advancing Slicer is a dedicated
  image-bump PR that runs full CI in isolation; upstream breakage is absorbed
  there, not scattered across contributors' PRs. This is the single place the
  two dev branches (Slicer main, Slicer-Liver `preview`) are reconciled, on
  our cadence.
- **Runtime smoke-check** in each job's verify stage: `PythonSlicer` asserts
  `slicer.util` carries `exit` + `VTKObservationMixin`, so a pinned-but-broken
  image fails fast with an ADR-cited message instead of as a downstream hang.
  Uses a library import (no event loop) so it can't itself hang.

## Files

- `.github/workflows/ci.yml` — both `image:` pins + two smoke-check steps.
- `Docs/adr/0030-ci-slicer-image-pinning.md` — the decision (Accepted).
- `Docs/index.md` — ADR-0030 added to the toctree.

## Trade-off

New Slicer features aren't picked up until someone bumps the pin (controlled
staleness over uncontrolled churn). Alternatives (digest pin, scheduled
auto-bump) are weighed in the ADR.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,14 +124,24 @@ jobs:
     # project-maintained image (Ubuntu 24.04 + Qt 6.9 + Slicer main
     # pre-built); recipe and auto-build workflow live in the
     # `ALive-Docker` repository (slicer-build-ubuntu2404/Dockerfile +
-    # .github/workflows/build-image.yml).  Rebuilt weekly to track
-    # Slicer main; pin to a `:main-<sha>` tag here if reproducible
-    # builds become important.
+    # .github/workflows/build-image.yml).
+    #
+    # PINNED to a specific Slicer-main build, NOT `:latest` — see
+    # Docs/adr/0030-ci-slicer-image-pinning.md.  `:latest` is a rolling
+    # tag rebuilt from Slicer main; a moving base means every PR builds
+    # against a different, non-reproducible Slicer and a bad upstream
+    # push lands on whichever PR happens to run during the window.  The
+    # `:main-<sha>` tag below names the exact Slicer commit; bump it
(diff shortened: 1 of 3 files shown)


RafaelPalomar (commented):
LGTM

PR #454, opened 2026-06-09
ENH: Unify Subject Hierarchy folder placement via vtkSlicerSubjectHierarchyFolders (T5.2-f)

RafaelPalomar:
## Summary for humans

Slicer-Liver placed its programmatically-created nodes under per-stage Subject-Hierarchy folders, but every node-creating module open-coded its own lookup / lazy-create / reparent dance — four copies that could drift. This adds a standalone wrapped C++ kit, `vtkSlicerSubjectHierarchyFolders`, that centralises that logic behind one static method and four folder-name accessors that are the single source of truth.

The kit links MRMLCore only (no Qt, no module-Logic dependency), so it is a ctkTest-reachable C++ low-level utility, and it is Python-wrapped (`vtkSlicerSubjectHierarchyFoldersPython`) so the Python caller links the exact compiled method the C++ callers do. Its name drops the `Liver` prefix per the closed-vocabulary class-naming convention.

Three consumers are wired to it:

- **VascularTerritories** — the inline `OnMRMLSceneNodeAdded` collector becomes a `CollectUnderFolder` call (behaviour preserved, proven by a characterization test).
- **LiverResections** — `OnMRMLSceneNodeAdded` collects the surgeon-facing wrapper `vtkMRMLLiverResectionNode` under `Resections`; the hidden `SetHideFromEditors(true)` Bezier carriers stay unparented.
- **LiverSegmentation** — `_collectUnderAnatomyFolder` delegates to the wrapped utility, keeping the lazy / headless-safe contract.

Scope: 3 active modules wired now; **LiverVolumetry deferred** — it creates no nodes today, so the `Volumetry` name + accessor exist but no consumer calls it yet. An ADR-0023 amendment records the convention and the ADR-0004 reasoned exception (wrapped C++ rather than Python-by-default, so C++ and Python callers share one binary-identical implementation).

## ADR references

1. **ADR-0023** — amended with a new "Subject Hierarchy management convention" section (closed four-folder vocabulary, shared utility as single source, named-constant accessors, node placement convention).
2. **ADR-0004** — reasoned exception recorded in the amendment: wrapped C++, not Python-by-default.
3. **ADR-0014** — wrapper-vs-carrier split: only the wrapper is collected, not the hidden carriers.
4. **ADR-0003 / ADR-0008 §2** — the utility is a C++ low-level (ctkTest) target, MRMLCore-only.
5. **ADR-0027** — test-first: the RED utility-core test predates this implementation commit; conformance points tagged `[test]` / `[review]` / `[future]`.

## Conformance

| Invariant | Proven by |
| --- | --- |
| Folder-name accessors return the four names verbatim | `vtkSlicerSubjectHierarchyFoldersTest1` (constants) — C++, verified green locally |
| Placement under a scene-root folder; lazy-create + reuse (one folder, not two) | `vtkSlicerSubjectHierarchyFoldersTest1` (placement + reuse) — C++, verified green locally |
| Null scene / node / folder name return false, no side effect | `vtkSlicerSubjectHierarchyFoldersTest1` (null-arg safety) — C++, verified green locally |
| VascularTerritories collection observably-identical pre/post rewrite | `vtkSlicerVascularTerritoriesLogicSubjectHierarchyCharacterizationTest` — C++, verified green locally (equivalence oracle) |
| LiverResections wrapper lands under `Resections`; hidden Bezier carrier does not | `test_resections_subject_hierarchy_collection.py` — launched-Slicer, CI proves |
| Wrapped utility importable + callable from Python; idempotent Anatomy placement | `test_liversegmentation_subject_hierarchy_folders_reachability.py` — launched-Slicer, CI proves |
| Consumers reference accessors, never literals | `[review]` |
| LiverVolumetry adoption | `[future]` — deferred |

**Verified locally (C++ CTests, green):** `vtkSlicerSubjectHierarchyFoldersTest1`, `vtkSlicerVascularTerritoriesLogicSubjectHierarchyCharacterizationTest`.

**CI must prove (launched-Slicer pytest):** the two `*subject_hierarchy*` pytest entries need a live `qSlicerApplication` to reach `slicer.mrmlScene` and the registered module logic / the wrapped kit on the interpreter; under bare CTest they skip cleanly (verified locally — both skip, not error).

## UX impact

N/A — non-UI change. Subject-Hierarchy folder organisation is surgeon-visible in the Data module's hierarchy view, but this PR only unifies the existing placement mechanism; the folder set, names, and node parenting the surgeon sees are unchanged from the behaviour ADR-0023 already specified.

<details>

### Why a wrapped C++ kit and not a Python helper

ADR-0004 makes Python the default for orchestration glue. This utility is a reasoned exception because the collection fires from two C++ scene observers (`vtkSlicerVascularTerritoriesLogic::OnMRMLSceneNodeAdded`, `vtkSlicerLiverResectionsLogic::OnMRMLSceneNodeAdded`) *and* from a Python call site (`LiverSegmentation._collectUnderAnatomyFolder`). Only a wrapped C++ implementation gives all three one binary-identical code path. A Python helper would be unreachable from C++, forcing either a second C++ copy (the drift this convention exists to prevent) or a C++→Python round-trip from inside a scene observer.

### Headless / null-SH contract

`CollectUnderFolder` resolves the Subject-Hierarchy node and early-returns `false` when none is available, so a missing SH plugin in a headless context never breaks the caller's node creation. Null scene / node / folder name also return `false` cleanly. The C++ test pins these as crash-free, false-returning observables.

### The bare-CTest skip fix for the LiverResections test

`LiverResections/Testing/Python/` carries an `__init__.py` (it is a Python package), so pytest's prepend-import walks up to `LiverResections/Testing` as `sys.path[0]` and a bare `from conftest import ...` resolves to the cross-module root `Testing/Python/conftest.py` (which does not re-export the underscore guard names) — an `ImportError` rather than a clean skip. This was a pre-existing scaffold defect (reproducible on the RED test commit, independent of this PR's production code). The fix imports the guards from their canonical home `slicer_pytest_support` (always on `sys.path` via the `pythonpath` ini option), which both conftests themselves use. No invariant is weakened. The sibling LiverSegmentation test has no `__init__.py` and was already skipping cleanly.

### Local build / test environment

Built C++ Debug against the preview Slicer + LayerDisplayableManager builds; ran the two C++ CTests directly (green) and confirmed the two pytest entries skip cleanly under bare CTest. The launched-Slicer pytest rows are CI's responsibility.
</details>


RafaelPalomar (commented):
LGTM

PR #458, opened 2026-06-11
ENH: Add weak target-model reference to vtkMRMLBezierSurfaceNode (ADR-0014 §1)

RafaelPalomar:
## Summary for humans

Adds a typed **weak** node reference on `vtkMRMLBezierSurfaceNode` to its
target organ (liver) model node, per ADR-0014 §1. The reference is
non-owning and non-observing: it is registered in the node constructor via
`AddNodeReferenceRole` with no events array and no content-modified
observation — the same shape as the existing `geometry`-role precedent on
`vtkMRMLResectionPlanNode`. Mutating the target therefore does not advance
the surface node's MTime.

The change is the node API only:

- `static const char* GetTargetReferenceRole()` — single source of truth
  for the role string (closed-vocabulary naming, no `Liver` prefix).
- `vtkMRMLModelNode* GetTargetModelNode()` — typed resolver.
- `void SetAndObserveTargetModelNode(vtkMRMLModelNode*)` — wire / clear.

These mirror the `GetGeometryNode` / `SetAndObserveGeometryNode` accessor
pair on the plan node. Wiring the Init-mode Representations to set and
consume the reference is the downstream T2 ring-extraction work
(`TODO(T2-target-mesh-weakref)`) and is intentionally out of scope here, so
this lands as a clean prerequisite.

## ADR references

1. ADR-0014 §1 — Bezier surface node weak-refs the target organ model node.

## Conformance

Invariants honoured, with the four sub-tests in
`LiverResections/MRML/Testing/Cxx/vtkMRMLBezierSurfaceNodeTest1.cxx` that
prove each:

- `testTargetReferenceSetGet` — role accessor is a stable non-empty string;
  the typed setter/getter round-trip and the reference is reachable under
  the canonical role string.
- `testTargetReferenceWeakSemantics` — mutating the target (`Modified()` +
  `SetName`) does **not** advance the surface node's MTime (the weak,
  non-observing contract).
- `testTargetReferenceSceneRoundTrip` — the reference persists through a
  `.mrml` scene Commit/Connect cycle and re-resolves by ID, and survives an
  in-memory `Copy`.
- `testTargetReferenceAbsentReturnsNull` — absent reference returns nullptr
  cleanly; clearing via nullptr and removing the target from the scene both
  return to the absent state.

Local CTest result (Debug build, runs `WITH_VTK_DEBUG_LEAKS_CHECK`, so
green also proves no leak):

```
1/1 Test #9: vtkMRMLBezierSurfaceNodeTest1 ....   Passed    0.44 sec
100% tests passed, 0 tests failed out of 1
```

## UX impact

N/A — non-UI change (MRML node API only).

<details>
<summary>Agent context (long form)</summary>

The implementation rides the standard MRML node-reference machinery rather
than introducing any special case: `AddNodeReferenceRole(role, role)` with
the two-argument form (reference role + MRML attribute name, no events
array, `observeContentModifiedEvents` left at its non-observing default)
exactly mirrors the `geometry` role on `vtkMRMLResectionPlanNode`. The typed
accessors are thin wrappers over `GetNodeReference` /
`SetAndObserveNodeReferenceID` + `SafeDownCast`, so scene-XML round-trip and
`CopyContent`/`Copy` carry the reference for free once the role is
registered.

The four committed RED sub-tests each constructed a bare `vtkMRMLScene` and
then called `scene->RegisterNodeClass(vtkMRMLModelNode)`. `vtkMRMLScene`
already registers `vtkMRMLModelNode` at construction, so the
re-registration emitted five "Tag Model has already been registered"
warnings that the module test harness's error-output check counted as a
failure. Those redundant registration lines were scaffolding, not
assertions; dropping them aligns the new sub-tests with the pre-existing
passing tests (which register only the module classes the scene does not
already know) and changes no assertion.

</details>

### DIFF
--- a/LiverResections/MRML/Testing/Cxx/vtkMRMLBezierSurfaceNodeTest1.cxx
+++ b/LiverResections/MRML/Testing/Cxx/vtkMRMLBezierSurfaceNodeTest1.cxx
@@ -1467,7 +1467,6 @@ int testTargetReferenceSetGet()
 {
   vtkNew<vtkMRMLScene> scene;
   scene->RegisterNodeClass(vtkSmartPointer<vtkMRMLBezierSurfaceNode>::New());
-  scene->RegisterNodeClass(vtkSmartPointer<vtkMRMLModelNode>::New());
 
   vtkNew<vtkMRMLBezierSurfaceNode> surface;
   vtkNew<vtkMRMLModelNode> target;
@@ -1517,7 +1516,6 @@ int testTargetReferenceWeakSemantics()
 {
   vtkNew<vtkMRMLScene> scene;
   scene->RegisterNodeClass(vtkSmartPointer<vtkMRMLBezierSurfaceNode>::New());
-  scene->RegisterNodeClass(vtkSmartPointer<vtkMRMLModelNode>::New());
 
   vtkNew<vtkMRMLBezierSurfaceNode> surface;
   vtkNew<vtkMRMLModelNode> target;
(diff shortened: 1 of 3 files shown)


RafaelPalomar (commented):
LGTM

PR #459, opened 2026-06-11
ENH: Wire T2 ring extraction on commit + quadric SSOT (Stack 4)

RafaelPalomar:
## Summary

Wire the unblocked slice of the T2 Stack-4 ring-extraction design constraints (#337), now that the `vtkMRMLBezierSurfaceNode` target weakref (#458) has landed. The three #337 constraints are carried as acceptance criteria (no new ADR).

## Changes

1. **Quadric SSOT (Constraints 1+2 — parameter consistency + granularity).** Hoist the spheroid implicit's `a0…a9` expansion into a single source of truth, `vtkLiverSpheroidRingExtractor::ComputeQuadricCoefficients(center, rx, ry, rz, a[10])`. `RequestData` now feeds its `vtkQuadric` from it (behaviour-identical refactor), and the Stack-4 parameter→shader adapter transcribes the same implicit from here — eliminating the off-by-2 transcription-drift class on the linear terms.
2. **On-commit-not-per-drag (Constraint 3, ADR-0019).** Add `commit()` / `_pending_extraction` / `_run_ring_extraction` to `LiverBezierSurfacePipeline`. Per-drag `UpdatePipeline` ticks in `Init` only mark extraction pending (per-frame visual feedback is the shader's job); the discrete CPU ring extraction runs exactly once on the irreversible `Init → Planning` transition.
3. **Weakref consumption (ADR-0014 §1).** The Init Representations' `TODO(T2-target-mesh-weakref)` sites resolve the target mesh via `vtkMRMLBezierSurfaceNode::GetTargetModelNode()`.

## Conformance

| Constraint | Test |
|---|---|
| 1+2 implicit SSOT + `(1/8)·max\|F″\|·h²` bound | `vtkLiverSpheroidRingExtractorConsistencyTest` (C++) |
| 3 on-commit-not-per-drag | `test_resections_ring_extraction_on_commit::test_extractor_not_invoked_per_drag_only_on_commit` (launched) |
| weakref consumption | `…::test_init_representation_reads_target_via_weakref` (launched) |

## Verified

- **Local C++ (via `Slicer --launch`):** `vtkLiverSpheroidRingExtractorConsistencyTest` → exit 0 (flipped from the 125-skip once the SSOT + `-DLIVER_SPHEROID_QUADRIC_SSOT` activated); the pre-existing `vtkLiverSpheroidRingExtractorTest1` bound test → exit 0 (SSOT refactor is behaviour-preserving).
[...]

Issue #460, opened 2026-06-11
pytest_launched harness can't reach loadable-module Python sub-packages (LiverResectionsLib) — launched tests skip


RafaelPalomar:
## Summary

The `pytest_launched` CTest harness adds each module's **source** directory to `--additional-module-paths`. That makes **scripted** modules' Python importable (e.g. #449's LiverSegmentation tests run fine), but a **loadable C++ module's** Python sub-package only lands on `sys.path` when its built `.so` loads — which never happens from a source-dir path. So launched tests that import a loadable module's Python sub-package **skip** instead of running.

## Concrete impact

- **#459** (T2 ring extraction, Stack 4): `test_resections_ring_extraction_on_commit.py` — both methods (`test_extractor_not_invoked_per_drag_only_on_commit`, `test_init_representation_reads_target_via_weakref`) **SKIP** in CI with `No module named 'LiverResectionsLib'`. The Pipeline is pure Python (a LayerDM Pipeline), so there's no C++-test escape hatch — these invariants are currently covered only by code review, not an executing test.
- Same root cause made **#454**'s `test_resections_subject_hierarchy_collection.py` skip (`'liverresections' module not registered`); that one had a C++-test alternative, tracked as #455.

## Ask

Make the launched harness reach loadable-module Python sub-packages so these tests execute:

- Point `pytest_launched`'s `--additional-module-paths` at the **built** loadable-module paths (`build/lib/Slicer-X.Y/qt-loadable-modules`) for C++ modules — not (only) the source dir — so loading them puts `LiverResectionsLib` (and peers) on `sys.path`; or
- otherwise ensure the loadable modules' staged Python sub-packages are importable in the launched interpreter.

Acceptance: the two #459 launched tests **execute (not skip)** and pass under `pytest_launched`; verify no other module's launched tests regress (scripted-module reachability must stay working). Apply the #449 launched-test discipline (clean `mrmlScene.Clear()` teardown, no in-process `sys.modules['slicer']` stubbing).

Milestone: v2.0.0 (these gate genuine execution of shipped v2.0 invariants). Follow-up to #459; related to #455.

RafaelPalomar:
## Diagnosis from the first harness-fix attempt (closed PR #461, branch `feature/launched-harness-loadable-paths` kept as reference)

[...]

PR #461, opened 2026-06-11
BUG: Reach loadable-module Python in pytest_launched harness (built module dirs)

RafaelPalomar:
## Summary

Fix the `pytest_launched` harness so it can reach **loadable-module Python sub-packages**, ending the class of silent skips where launched tests for C++ modules' Python (LayerDM Pipelines/Representations) never actually run.

## Root cause

The harness added each module's **source** dir to `--additional-module-paths`. That makes a *scripted* module's Python importable, but a *loadable* (C++) module's Python sub-package — e.g. `LiverResectionsLib` — is staged into the built `qt-scripted-modules` tree and only reaches `sys.path` when the module loads from a **built** path. So launched tests importing such a sub-package SKIPPED with `No module named 'LiverResectionsLib'`, silently under-verifying shipped behaviour.

## Fix

Point `--additional-module-paths` at the two built module dirs
(`${CMAKE_BINARY_DIR}/${Slicer_QTLOADABLEMODULES_LIB_DIR}` and `…QTSCRIPTEDMODULES…`) instead of the source dirs. These cover all six modules (scripted + loadable).

## Verification

- **Headless launched probe (local):** with the built dirs, all six modules register (`Liver`, `LiverResections`, `LiverMarkups`, `LiverVolumetry`, `VascularTerritories`, `LiverSegmentation` all `True`) and `import LiverResectionsLib.LiverBezierSurfacePipeline` succeeds — versus `No module named 'LiverResectionsLib'` from the source dirs.
- **Generated CTest command:** confirmed it now resolves to `.../build/lib/Slicer-5.11/qt-{loadable,scripted}-modules`.
- **CI proves the payoff:** the already-merged `test_resections_subject_hierarchy_collection.py` (from #454) currently SKIPS on `liverresections not registered`; with this change it should **execute** under `pytest_launched`. (Once #459 merges, its ring-extraction-on-commit + weakref launched tests also un-skip.)

## Notes

Scripted-module reachability (e.g. LiverSegmentation's #449 tests) is preserved — the built `qt-scripted-modules` dir carries them too (verified by the all-modules-register probe). No source-dir dependency remains.

---

### DIFF
--- a/Liver/Testing/Python/CMakeLists.txt
+++ b/Liver/Testing/Python/CMakeLists.txt
@@ -118,20 +118,24 @@ else()
   return()
 endif()
 
-# Full in-tree module set for ``--additional-module-paths``.  Each entry is a
-# module source dir whose scripted/loadable module Slicer must discover so the
-# launched widget-level tests can import it.  Keep in sync with the top-level
-# CMakeLists.txt module list.
+# Module paths for ``--additional-module-paths``: the BUILT module
+# directories, NOT the in-tree source dirs.
+#
+# A module source dir only makes a *scripted* module's Python importable.
+# A *loadable* (C++) module's Python sub-package (e.g. ``LiverResectionsLib``,
+# which carries the LayerDM Pipelines + Representations) is staged into the
+# built ``qt-scripted-modules`` tree and only lands on ``sys.path`` when the
+# module loads from a built path — never from the source dir.  Pointing the
+# launched harness at the source dirs therefore made every test that imports
+# a loadable module's Python sub-package SKIP with ``No module named
+# 'LiverResectionsLib'`` (e.g. the resection-plan SH-collection and ring-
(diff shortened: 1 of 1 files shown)


RafaelPalomar:
Closing without merging — the built-module-path change here is correct but **insufficient on its own**, and turning it on surfaced a deeper two-part problem (full diagnosis on #460). Keeping the branch `feature/launched-harness-loadable-paths` as reference for whoever picks up #460.

Summary: pointing the harness at the built module dirs DID make all six modules register and `LiverResectionsLib` reachable on-path, but (1) CI's launched base Slicer doesn't load the SlicerLayerDM extension, so `LiverResectionsLib`'s hard `from LayerDMLib import …` (ADR-0013 §5) raises and the target SH/ring tests still skip; and (2) it un-skipped 5 stale `test_bezier_surface_node.py` tests that now fail on API drift. Both need addressing before this lands green.

Issue #455, opened 2026-06-11
Add executing C++ logic test for LiverResections Subject-Hierarchy wiring (follow-up to #454)


RafaelPalomar:
## Context

#454 (T5.2-f, closed #412) wired LiverResections Subject-Hierarchy collection: in `vtkSlicerLiverResectionsLogic::OnMRMLSceneNodeAdded`, after the `vtkMRMLLiverResectionNode::SafeDownCast` guard, the wrapper node is collected under the "Resections" folder via `vtkSlicerSubjectHierarchyFolders::CollectUnderFolder` (hidden Bezier/contour carriers are not reparented).

That wiring currently has **no executing test**. The launched-Slicer pytest written for it (`LiverResections/Testing/Python/test_resections_subject_hierarchy_collection.py`) **skips** in CI and headless because a loadable C++ module does not load from the `pytest_launched` source-dir `--additional-module-paths` (and the pre-existing headless dependency-load quirk). It's covered by code-inspection + the green `vtkSlicerSubjectHierarchyFoldersTest1` utility test, but nothing green fires the observer.

## Ask

Add a **C++ logic test** that actually runs in CTest — mirroring the pattern of `VascularTerritories/Logic/Testing/Cxx/vtkSlicerVascularTerritoriesLogicSubjectHierarchyCharacterizationTest.cxx`:

- instantiate `vtkSlicerLiverResectionsLogic`, set its scene, register node classes;
- create/add a `vtkMRMLLiverResectionNode` (wrapper) plus a hidden Bezier/contour carrier;
- drive the `NodeAdded` path;
- assert the wrapper lands under a scene-root "Resections" folder and the hidden carrier is **not** reparented;
- leak-checked (`WITH_VTK_DEBUG_LEAKS_CHECK`), like the existing C++ tests.

This converts the one green-but-skipping invariant from #454 into a genuinely-executing test. The launched-pytest can stay (it runs when a full Slicer loads the module), but the C++ logic test is the one that proves it in CI.

Milestone: v2.0.0. Quality follow-up to #454 / #412.

RafaelPalomar:
Delivered by #462 (merged): executing C++ logic test for the LiverResections SH wiring — verified locally exit 0 (also confirms the #454 OnMRMLSceneNodeAdded → CollectUnderFolder wiring is correct, not just present). Closes the #454 inspection-only gap. The launched-pytest counterpart remains tracked under #460.

PR #462, opened 2026-06-11
ENH: Add executing C++ logic test for LiverResections Subject-Hierarchy wiring (#455)

RafaelPalomar:
## Summary

Add an **executing** C++ logic test for the LiverResections Subject-Hierarchy wiring (follow-up to #454, closing the inspection-only gap noted there).

## Why

#454 wired resection-plan SH collection in `vtkSlicerLiverResectionsLogic::OnMRMLSceneNodeAdded` (→ `vtkSlicerSubjectHierarchyFolders::CollectUnderFolder`, ADR-0023 §"Subject Hierarchy management convention"). Its launched-pytest counterpart **skips** on the harness loadable-module reachability gap (#460), so the wiring was covered only by code review. This adds a test that actually runs.

## The test

`vtkSlicerLiverResectionsLogicSubjectHierarchyCharacterizationTest` (mirrors the `VascularTerritories` precedent): instantiates the logic, sets the scene, drives `NodeAdded`, and asserts:
- the wrapper `vtkMRMLLiverResectionNode` lands under a scene-root **"Resections"** folder (folder name via `GetResectionsFolderName()`, not a literal);
- a hidden carrier is **not** reparented (wrapper-vs-carrier, ADR-0014).

Runs in CTest (no launched-harness dependency); leak-checked (`WITH_VTK_DEBUG_LEAKS_CHECK`).

## Verified

Built locally and run via `Slicer --launch` → **exit 0** ("completed successfully"). This green run also confirms the #454 SH wiring is correct, not just present.

---

### DIFF
--- a/LiverResections/Logic/CMakeLists.txt
+++ b/LiverResections/Logic/CMakeLists.txt
@@ -33,3 +33,7 @@ SlicerMacroBuildModuleLogic(
   SRCS ${${KIT}_SRCS}
   TARGET_LIBRARIES ${${KIT}_TARGET_LIBRARIES}
   )
+
+if(BUILD_TESTING)
+  add_subdirectory(Testing)
+endif()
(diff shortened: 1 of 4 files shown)


RafaelPalomar (commented):
LGTM

PR #473, opened 2026-06-12
BUG: Make volume-bounded-by-resection work in v2 (retarget volumetry to the parametric Bezier carrier)

RafaelPalomar:
## Summary for humans

In v2 the "compute volume bounded by a resection" feature was inert: the
resection-target combo in LiverVolumetry filtered on the legacy
`vtkMRMLLiverResectionNode` (never instantiated in v2), so it stayed empty,
and even if a node reached the logic, `vtkLiverVolumetryLogic` downcast each
collection item to the disjoint v1 Markups type
`vtkMRMLMarkupsBezierSurfaceNode`. The v2 resection-plan geometry carrier is
`vtkMRMLBezierSurfaceNode` (the parametric hierarchy), so the `SafeDownCast`
returned `nullptr`, the Bezier boundary was never projected onto the
target-segment label map, and the volume was computed with no resection
boundary at all.

This PR fixes the carrier/Markups type mismatch end to end:

- **C++ logic retarget**: the downcast in `GetResectionsProjectionITKImage`
  now targets `vtkMRMLBezierSurfaceNode`; `GenerateBezierSurface` and `GetRes`
  take the parametric carrier and re-source the 16 control points from its
  flat row-major `GetControlGrid()` instead of the Markups
  `GetNthControlPointPosition` API. Non-carrier collection items are skipped
  before the grid is read.
- **Scripted-module retarget**: the resection-target combo now filters on the
  v2 surgeon-facing wrapper `vtkMRMLResectionPlanNode`, and
  `getResectionNodes()` collects each checked plan's geometry carrier via
  `GetGeometryNode()` (guarding plans without geometry).

This removes the last `GetBezierFromResection` caller, unblocking T2.7-2b.

## ADR references

1. ADR-0014 §"Fourth layer" — wrapper-vs-carrier geometry split; the
   resection-plan wrapper references a parametric `vtkMRMLBezierSurfaceNode`
   carrier whose control polygon lives in a flat control grid.

## Conformance

- **Invariant**: the LiverVolumetry surface-generation path, given a v2
  parametric carrier with 16 control points set via `SetControlGrid`, reads
  those points and produces a non-empty projected Bezier surface.
- **Tests**: `vtkLiverVolumetryLogicResectionCarrierTest`
  - `testProductionDowncastDropsCarrier` — pins the root cause (the v1 Markups
    `SafeDownCast` of a v2 carrier yields `nullptr`); stays green.
  - `testCarrierProducesNonEmptySurface` — was RED, now green: the
    carrier-accepting seam emits a non-empty surface.

  Local run (Slicer-launched CxxTest driver): both cases exit 0; `ctest`
  reports `100% tests passed, 0 tests failed out of 1`.

## UX impact

The resection-target combo in LiverVolumetry now populates with the v2
resection-plan nodes (previously always empty), so a surgeon can select one
or more resections as the volume boundary and the computed volume reflects
that boundary. No layout or control changes; the existing combo simply lists
the correct node type.

<details>

Root cause was a node-family mismatch introduced by the v1→v2 resection
rework. The v1 path stored resection geometry as a Markups node
(`vtkMRMLMarkupsBezierSurfaceNode : vtkMRMLMarkupsNode`) and read control
points through the Markups control-point API. v2 moved the geometry to a
parametric carrier (`vtkMRMLBezierSurfaceNode :
vtkMRMLAbstractParametricSurfaceNode`) whose 16 control points live in a flat
row-major grid of `3 * Rows * Cols` doubles (a 4x4 grid = 48 doubles), read
via `GetControlGrid()`. The two Bezier node families are disjoint subtrees of
`vtkMRMLNode`, so the unretargeted `SafeDownCast` silently dropped every v2
carrier to `nullptr` and the projection step ran on an empty surface source.

The fix re-sources control points by index `i -> grid[i*3 + {0,1,2}]` in both
consumers (`GenerateBezierSurface` for the full 4x4 surface, `GetRes` for the
two boundary curves at control indices `{3,6,9,12}` and `{0,5,10,15}`). The
`GenerateBezierSurface` population guard switched from
`GetNumberOfControlPoints() == 16` to `GetControlGridLength() == 48`, the
parametric-carrier equivalent for the fixed 4x4 `vtkBezierSurfaceSource`. A
loop guard skips any collection item that fails the carrier downcast before
`GetRes` dereferences the raw grid pointer.

The logic library already links `vtkSlicerLiverResectionsModuleMRML`, so no
new link dependency was needed.

</details>

### DIFF
--- a/LiverVolumetry/LiverVolumetry.py
+++ b/LiverVolumetry/LiverVolumetry.py
@@ -184,10 +184,15 @@ class LiverVolumetryWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
   def getResectionNodes(self):
     resectionNodes = vtk.vtkCollection()
     if not self.ui.ResectionTargetNodeComboBox.noneChecked():
-      lvLogic = slicer.modules.liverresections.logic()
       checkedNodes = self.ui.ResectionTargetNodeComboBox.checkedNodes()
       for i in checkedNodes:
-        bs = lvLogic.GetBezierFromResection(i)
+        # The checked nodes are now vtkMRMLResectionPlanNode wrappers; their
+        # parametric geometry carrier (vtkMRMLBezierSurfaceNode) is the
+        # boundary surface used for the volume computation
+        # (ADR-0014 wrapper-vs-carrier split).
+        bs = i.GetGeometryNode()
+        if bs is None:
+          continue
         resectionNodes.AddItem(bs)
     else:
       resectionNodes = None
--- a/LiverVolumetry/Logic/CMakeLists.txt
+++ b/LiverVolumetry/Logic/CMakeLists.txt
@@ -33,3 +33,7 @@ SlicerMacroBuildModuleLogic(
   SRCS ${${KIT}_SRCS}
   TARGET_LIBRARIES ${${KIT}_TARGET_LIBRARIES}
 )
+
+if(BUILD_TESTING)
+  add_subdirectory(Testing)
+endif()
--- a/LiverVolumetry/Logic/Testing/CMakeLists.txt
+++ b/LiverVolumetry/Logic/Testing/CMakeLists.txt
@@ -1,0 +1,1 @@
+add_subdirectory(Cxx)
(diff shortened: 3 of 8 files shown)


RafaelPalomar (commented):
LGTM

PR #478, opened 2026-06-12
ENH: Skip the visual replay fast on software GL

RafaelPalomar:
## Summary for humans

The LiverResections visual-regression mirror tests run on CI under
`xvfb` + `LIBGL_ALWAYS_SOFTWARE` (llvmpipe). On that software-GL stack
the bezier distance-map render is unreliable and hangs until each
scenario hits its 180s CTest `TIMEOUT` — roughly ten minutes of dead
time on every CI run, repo-wide.

This adds a fast renderer probe at the top of `replay_test.py`: bring
up a throwaway offscreen `vtkRenderWindow`, force context creation with
an *empty* render, and read `ReportCapabilities()`. An empty render
touches only context + framebuffer setup, **not** the bezier mapper
that actually hangs, so the probe stays cheap even on llvmpipe. When
the renderer string matches a software rasteriser (`llvmpipe`,
`softpipe`, `swrast`, ...), or when no context can be created at all,
the driver prints a greppable `[skip]` line naming the renderer and
returns 0 — before the hanging render. Real GPUs proceed exactly as
before.

The root-cause render hang itself is out of scope here and tracked
separately under the visual-regression hardening work.

## ADR references

1. ADR-0020 §"Rollout plan" — visual-regression capture/replay rollout
   and hardening; the render-hang root cause lives here.
2. ADR-0003 — characterisation tests gate behaviour-changing PRs.
3. ADR-0008 §"observability" — pure-Python helpers carry self-tests.

## Conformance

- **Software-GL skip is distinct from the missing-baseline skip** — it
  is decided on the *renderer*, before and independently of
  `_has_baseline`. Proven by
  `test_software_gl_skip.py::test_skip_reason_independent_of_baseline`.
- **Renderer classification** (software vs real GPU) is a pure,
  unit-tested helper. Proven by
  `test_is_software_renderer_matches_software_stacks` /
  `_passes_real_gpus`.
- **A failed probe skips rather than proceeds** (no context → cannot
  render). Proven by `test_skip_reason_set_when_no_context`.
- **Probe never crashes the run** — the context probe is wrapped in
  try/except, returning `None` (→ skip) on any failure.

## UX impact

N/A — non-UI change (CI test driver only).

<details>
<summary>Agent context (long form)</summary>

### Why an empty render forces the context

A bare offscreen `vtkRenderWindow` creates its GL context lazily on the
first `Render()`. Before that, `ReportCapabilities()` returns only
`display id not set` — verified offscreen on Mesa — so a capability
query alone is insufficient to classify the renderer. An empty
`Render()` (no renderer, no actors, no mapper) is the cheapest way to
force context creation and populate the GL strings.

### Local verification

Run on an AMD Radeon (radeonsi) host, Slicer launcher under
`--no-main-window`:

- **Real GL** (`DISPLAY=:0`): probe reports
  `OpenGL renderer string: AMD Radeon Graphics (radeonsi, ...)`,
  classified non-software → decision PROCEED. The full replay then
  proceeds past the probe into the render.
- **Forced software GL** (`LIBGL_ALWAYS_SOFTWARE=1 GALLIUM_DRIVER=llvmpipe`):
  the empty-render probe comes up under llvmpipe without hanging,
  reports `llvmpipe (LLVM 18.1.8, 256 bits)`, classified software →
  the full `replay_test.py` prints the `[skip]` line and returns 0 in
  ~5s (wrapped in `timeout 90`, no `rc=124`). This is the path that
  previously consumed the 180s timeout.

The probe itself does **not** hang on llvmpipe — the specific risk the
task flagged — because it never reaches the bezier mapper.

### Test additions

`test_harness/test_software_gl_skip.py` (13 cases): renderer-string
classification both directions, render-vs-skip decision + message
shape, no-context skip, baseline-independence, and renderer-token
formatting. Pure-Python, no GL context, runs under the existing
`LiverResections_harness_self_tests` CTest entry. Full harness suite:
36 passed.

</details>

### DIFF
--- a/LiverResections/Testing/Python/replay_test.py
+++ b/LiverResections/Testing/Python/replay_test.py
@@ -46,6 +46,24 @@ import sys
 # catching real visual regressions.
 DEFAULT_TOLERANCE = 0.15
 
+# Substrings that mark a GL renderer string as a software rasteriser.
+# When the replay runs on one of these (e.g. CI's xvfb + llvmpipe
+# stack), the bezier distance-map render is unreliable and has been
+# observed to hang to the CTest TIMEOUT (see the visual-regression
+# hardening issue tracked in ADR-0020 §"Rollout plan").  We detect
+# this up front and skip BEFORE the hanging render rather than burning
+# the full per-test timeout.  Matched case-insensitively against the
+# renderer + version strings reported by ``vtkRenderWindow``.
+SOFTWARE_GL_MARKERS = (
+    "llvmpipe",
+    "softpipe",
+    "swrast",
+    "software rasterizer",
+    "software rasteriser",
+    "gallium llvmpipe",
+    "mesa offscreen",
+)
+
 
 def _parse_argv() -> argparse.Namespace:
     """Slice out the ``--`` separated args; same convention as the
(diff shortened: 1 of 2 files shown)


RafaelPalomar (commented):
LGTM

PR #479, opened 2026-06-17
BUG: Activate Bezier distance-map sampler3D per render (#475)

### LINKED ISSUE #475: BUG: Bezier resection distance-map render (glTexImage3D 3D-texture path) aborts in the current build
## Symptom

Rendering a `vtkMRMLMarkupsBezierSurfaceNode` with a distance-map volume attached **hard-aborts** the Slicer process (`SlicerApp-real exit abnormally`) during `forceRender()`. Reproduced on `DISPLAY=:0` (real GL) on `curie` with the current `preview` build, and independently in a headless agent context.

## Isolation

| scene state | 2nd `forceRender()` |
|---|---|
| bezier visible, no texture | OK |
| visible + `TextureNumComps=4`, **no** distance map | OK |
| visible + texture + **distance map attached** | **process aborts** |

- A trivial model scene (sphere) renders fine on the same host/session, so general GL works.
- Reproduces with a generic 10×10×10 4-component `vtkMRMLScalarVolumeNode` — independent of how the scene is constructed (legacy or v2 path).
- The abort is in the relocated `vtkOpenGLBezierResectionPolyDataMapper`'s 3D distance-map texture upload / sampler path (`glTexImage3D` with a multi-component volume).

## Why it matters / why it wasn't caught

This is a **core v2.0 render path** (the resection surface tinted by the parenchyma distance map). The visual-regression mirror tests that would catch it are **soft-gated / skipped in CI** (offscreen software GL hangs, see #474), so a render regression here is invisible to CI — exactly the verification-escape gap. The blessed `BezierSurface4x4{Planning,Confirmed}` baselines were captured 2026-05-20 (the render worked then); it aborts now, so the baselines **cannot currently be reproduced on any available host**, which blocks pixel-diff verification of any scenario touching this path.

## Prime suspect

The mapper relocation (LiverMarkups → `LiverResections/VTKWidgets`, #468, ~2026-06-11) is the leading candidate for a regression — it moved this mapper and the soft-gated visual tests would not have flagged a render break. Could also be a VTK/GL-version interaction in the current build. Needs a bisect: does the distance-map render abort at #468's parent vs HEAD?

## Impact / relation

- Blocks pixel-parity verification for the T2.7-2b visual-scenario retarget (committed by-construction; baselines untouched).
- Directly motivates #474 (get a real headless/EGL render gate so this path is exercised in CI).

## Next steps

1. Reproduce + capture the actual abort (stack/GL error) under a debugger.
2. Bisect across #468 to confirm regression-vs-environment.
3. If a code regression: fix the 3D-texture upload in `vtkOpenGLBezierResectionPolyDataMapper`. If environmental: pin the working GL/VTK config + document the capture host.

RafaelPalomar:
## Summary for humans

Closes #475 (candidate fix — **pixel/abort verification pending a maintainer `:0` GPU render**).

The offscreen-render abort on the *second* `forceRender()` of a Bezier
surface carrying a 4-component float distance-map is a latent draw-time
texture-binding bug, not a relocation regression. The 3D distance-map
texture was `Bind()`/`Deactivate()`-ed exactly once at creation; at draw
time the mapper only set the `sampler3D distanceTexture` uniform to a
hardcoded texture unit 0 and relied on that stale one-time bind. Nothing
re-activated the texture object onto a unit per render, so after VTK churns
texture-unit bindings for other actors a later render samples a `sampler3D`
with no live `GL_TEXTURE_3D` bound to its unit — which strict drivers fault
on during the draw.

The fix threads the texture object from the representation to the mapper,
`Activate()`s it at draw time, binds the sampler to the unit it actually
lands on, and `Deactivate()`s it after the draw — mirroring how the base
`vtkOpenGLPolyDataMapper` manages its own textures. Surgical correctness
patch on the v2.0 mapper; the GPU-tessellation rewrite is left for v2.1.

This is **test-first**: a characterisation test renders the 4-component
distance-map scenario and `forceRender()`s twice, pinning "the second render
must not abort". It is RED-by-design on the affected stacks and skips up
front on software GL (the verdict is on a GPU display).

### Maintainer verification (required before Ready)

1. Rebuild the `LiverResections` + `LiverMarkups` VTKWidgets libraries.
2. On a GPU-backed display, run the characterisation test:
   `Slicer --no-main-window --no-splash --additional-module-paths <repo>/LiverResections --python-script <repo>/LiverResections/Testing/Python/distancemap_double_render_test.py`
   (or via CTest: `ctest -R LiverResections_distance_map_double_render`).
   Expected: `[pass] Bezier 4-component distance-map survived two forceRender() calls` — **abort gone**.
3. Re-run the existing `LiverBezierVisualTest_BezierSurface4x4Planning`
   replay on the same display and confirm the render reproduces / pixels
   match the baseline (re-enables baseline reproduction for T3 / 2b).

### If H1 does not resolve it (fallback)

The keystone analysis ranks a second hypothesis (H2): a `RGBA32F` 3D texture
with a `Linear` min/mag filter faulting on linear-filtered float-3D sampling
on some drivers. Fallback: switch the distance-map texture Min/Mag filter to
`Nearest` in `CreateAndTransferDistanceMapTexture`. The gdb recipe in the
keystone plan distinguishes H1 (fault at the draw / sampler fetch) from H2
(fault at the linear-filtered sample) — `bt` at the abort lands in the GL
draw, not in `CreateSeq3DFromRaw`, if H1 is correct.

## ADR references

1. ADR-0003 §"Decision" — characterisation test pins behaviour before the fix.
2. ADR-0020 §"Rollout plan" §7 — v2.0 Bezier mapper is ported, not rewritten;
   this is a surgical correctness patch, not modernisation.

## Conformance

- **Invariant: a second offscreen render of the 4-component distance-map
  Bezier scenario must not abort.** Proven by
  `LiverResections_distance_map_double_render`
  (`distancemap_double_render_test.py`) — RED on the affected stacks today;
  green once the fix is confirmed on a GPU display. [test]
- **Invariant: the v2.0 mapper is not rewritten (ADR-0020).** The change is
  confined to threading + per-render activate/deactivate of an existing
  texture object. [review]

## UX impact

N/A — non-UI change (offscreen-render correctness in the Bezier resection
mapper; no user-facing surface, label, or workflow changes).

<details>

### Files changed

- `LiverResections/VTKWidgets/vtkOpenGLBezierResectionPolyDataMapper.{h,cxx}`
  — `SetDistanceMapTextureObject()` / `GetDistanceMapTextureObject()`
  accessors holding a `vtkSmartPointer<vtkTextureObject>`; `Activate()` +
  `SetUniformi("distanceTexture", GetTextureUnit())` in
  `SetMapperShaderParameters()`; `Deactivate()` in an overridden
  `RenderPieceFinish()`.
- `LiverMarkups/VTKWidgets/vtkSlicerBezierSurfaceRepresentation3D.cxx` —
  threads the texture object to the mapper at the end of
  `CreateAndTransferDistanceMapTexture()`.
- `LiverResections/Testing/Python/distancemap_double_render_test.py` +
  `CMakeLists.txt` registration (`LiverResections_distance_map_double_render`,
  outside the `SlicerLiver_ENABLE_VISUAL_TESTS` opt-in since it has no
  ExternalData baseline dependency).

### Why the relocation (#468) is exonerated

The `git mv` that relocated the mapper changed 0 body lines; the texture
upload helper (`vtkMultiTextureObjectHelper::CreateSeq3DFromRaw`) was not
moved. The bug is a pre-existing draw-time binding hazard, driver-agnostic
(reproduces on both llvmpipe and radeonsi per the keystone analysis), which
is why it is "our binding, not a vendor quirk".

### Build

Configured + built clean inside the worktree (`build/`, Debug) against the
local Slicer-build + SlicerLayerDM; both touched VTKWidgets libraries
compile with zero errors. Rendering could not be exercised in the
implementation environment (the bezier distance-map render aborts/hangs on
every host here) — hence the maintainer `:0` verification gate above.

</details>

### DIFF
--- a/LiverMarkups/VTKWidgets/vtkSlicerBezierSurfaceRepresentation3D.cxx
+++ b/LiverMarkups/VTKWidgets/vtkSlicerBezierSurfaceRepresentation3D.cxx
@@ -520,6 +520,10 @@ void vtkSlicerBezierSurfaceRepresentation3D::CreateAndTransferDistanceMapTexture
   {
     vtkWarningMacro("vtkSlicerBezierSurfaceRepresentation::CreateAndTransferDistanceMap:"
                     "There is no distance map node associated. Texture won't be generated.");
+    // Drop any previously-threaded texture so the mapper falls back to its
+    // no-distance-map path instead of sampling a stale 3D texture once the
+    // distance-map node is cleared (otherwise MRML state and GL state diverge).
+    this->BezierSurfaceResectionMapper->SetDistanceMapTextureObject(nullptr);
     return;
   }
 
@@ -528,6 +532,7 @@ void vtkSlicerBezierSurfaceRepresentation3D::CreateAndTransferDistanceMapTexture
   {
     vtkWarningMacro("vtkSlicerBezierSurfaceRepresentation::CreateAndTransferDistanceMap:"
                     "There is no image data in the specified scalar volume node.");
+    this->BezierSurfaceResectionMapper->SetDistanceMapTextureObject(nullptr);
     return;
   }
 
--- a/LiverMarkups/VTKWidgets/vtkSlicerBezierSurfaceRepresentation3D.cxx
+++ b/LiverMarkups/VTKWidgets/vtkSlicerBezierSurfaceRepresentation3D.cxx
@@ -539,6 +539,13 @@ void vtkSlicerBezierSurfaceRepresentation3D::CreateAndTransferDistanceMapTexture
   this->DistanceMapTexture->SetMagnificationFilter(vtkTextureObject::Linear);
   this->DistanceMapTexture->SetBorderColor(1000.0f, 1000.0f, 0.0f, 0.0f);
   this->DistanceMapTexture->CreateSeq3DFromRaw(dimensions[0], dimensions[1], dimensions[2], numComps, VTK_FLOAT, imageData->GetScalarPointer(), 0);
+
+  // Thread the texture object to the mapper so it is Activate()-ed onto a
+  // texture unit at draw time and the sampler3D uniform binds to that unit.
+  // The one-time Bind() above is not sufficient: on a later render the
+  // sampler3D would otherwise point at a unit with no live GL_TEXTURE_3D
+  // bound — the offscreen-render abort root cause (ADR-0003).
+  this->BezierSurfaceResectionMapper->SetDistanceMapTextureObject(this->DistanceMapTexture);
 }
 
 //----------------------------------------------------------------------
--- a/LiverResections/Testing/Python/CMakeLists.txt
+++ b/LiverResections/Testing/Python/CMakeLists.txt
@@ -192,6 +192,11 @@ set_tests_properties(py_distancemap_double_render_test PROPERTIES
   # Same timeout rationale as the replay entries: a stuck offscreen render
   # on a misdetected stack should fail fast rather than hold CI open.
   TIMEOUT 180
+  # The script exits 77 when it skips on a software-GL stack (CI's
+  # xvfb + llvmpipe).  Mark that as a CTest SKIP so a masked skip is visible
+  # as a skip on the dashboard, not an indistinguishable green pass; the real
+  # abort/no-abort verdict is on a GPU-backed display.
+  SKIP_RETURN_CODE 77
 )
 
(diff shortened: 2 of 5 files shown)


RafaelPalomar (commented):
LGTM

Issue #493, opened 2026-06-24
T2-mapper-wiring: drive the LayerDM resection Representation off the real relocated mapper


RafaelPalomar:
## T2-mapper-wiring — drive the LayerDM resection Representation off the real relocated mapper

**The literal ADR-0014 §3 mapper *relocation* is DONE** (commit `355ef69`: all four `vtkOpenGL*PolyDataMapper` classes live in `LiverResections/VTKWidgets/`, LiverMarkups consumes them by include path, `vtkOpenGLLiverResectionMappersRelocationTest1` link-tests green). The `TODO(T2-mapper-relocation)` tags scattered through the Python Representations are **misnamed** — they point at this remaining *wiring* work, not the completed relocation. This issue re-scopes + renames that work to **T2-mapper-wiring** and retires the stale tags.

### The gap
`LiverResectionsLib/Representations/BezierPlanningRepresentation.py` renders a **placeholder control-mesh** through a generic `vtkPolyDataMapper` (`_apply_data_node` builds the raw control grid, not the Bézier patch — `:391-400`). The **live production renderer** of the resection surface is still the v1 markups path: `vtkSlicerBezierSurfaceWidget → vtkSlicerBezierSurfaceRepresentation3D`, which builds the tessellated surface (`vtkBezierSurfaceSource` + `vtkPolyDataNormals` + the `uvCoords` array), computes the `RasToIjkMatrixT` / `IjkToTextureMatrixT` from the distance map, binds the distance-map 3D texture (`CreateAndTransferDistanceMapTexture` → `SetDistanceMapTextureObject`), and drives the real `vtkOpenGLBezierResectionPolyDataMapper`. So the remaining work is the **LayerDM render cutover for the Bézier resection surface** — porting that v1 setup into the LayerDM Representation.

### Why it matters
This is the bottleneck for the ADR-0025 locator: the consumer Representation (#489 second half) and Locator B/D need a real `vtkOpenGLBezierResectionPolyDataMapper` actually rendering the surface so the `uLocatorPosition`/`uLocatorRadius` uniforms (landed in #492) have visible effect.

### Proposed slices (invariant-test-first, ADR-0027)
[...]

PR #492, opened 2026-06-24
Locator C (shader half): uLocatorPosition / uLocatorRadius on the Bezier resection mapper

RafaelPalomar:
## Summary

The **C++ shader-uniform half** of Locator C (#489): `vtkOpenGLBezierResectionPolyDataMapper` gains `uLocatorPosition` / `uLocatorRadius` so the resection-surface shader can draw a locator marker at a world point (ADR-0025). Built on the now-merged Locator A node (#487).

## What it adds

- `vtkInternal`: `LocatorPosition[3]` (default `{0,0,0}`) + `LocatorRadius` (default `0` = off).
- Accessors mirroring the existing `Set/GetResectionMargin` family: `Get/SetLocatorPosition` (array + 3-scalar) and `Get/SetLocatorRadius`.
- Vertex shader: exports `vertexRASVSOutput = uShiftScale*vertexMC` (the RAS world position — `uShiftScale*vertexMC` is what `uRasToIjk` consumes downstream, verified against the existing texture-coord path).
- Fragment shader: `uLocatorPosition`/`uLocatorRadius` decls + a marker block that colours fragments within `uLocatorRadius` of `uLocatorPosition`. It is **added after** the existing corner-marker block (they coexist; corner-marker removal is #380/v2.1, per ADR-0025 Consequences).
- `SetMapperShaderParameters`: `IsUniformUsed`-gated binds for both uniforms.

## Testing

`vtkOpenGLBezierResectionPolyDataMapperLocatorTest1` — a GL-free C++ ctkTest pinning the ADR-0025 §Conformance accessor invariant (default off-state + round-trip of position/radius). **Passes** (1/1). The GLSL marker rendering itself is GPU/visual-regression and out of scope for the unit test.

## Deferred (consumer wiring)

The Python **consumer Representation** that would feed these uniforms from a `vtkMRMLLocatorNode` is **blocked on the T2-mapper-relocation** (ADR-0014 §3): `BezierPlanningRepresentation` still uses a placeholder `vtkPolyDataMapper`, not the real one, so the full node → consumer → `uLocatorPosition` chain cannot close yet. This PR lands the shader plumbing; #489 stays open for the consumer wiring once the relocation lands. See #489 for the scope split.

Part of #489 (ADR-0025 Locator).

---

🤖 Drafted by Claude (Anthropic Claude Opus 4.8) as an assistant to the maintainer.

Issue #489, opened 2026-06-21
Locator C: T2 consumer Representation + Bezier shader uniforms


RafaelPalomar:
Consumer Representation in the T2 host (`LiverBezierSurfacePipeline`) reads the `vtkMRMLLocatorNode` world point; extend `vtkOpenGLBezierResectionPolyDataMapper` with `uLocatorPosition` / `uLocatorRadius` uniforms via the existing `ReplaceShaderValues` / `SetMapperShaderParameters` mechanism (the `uResectionMargin` family seam). The locator marker **displaces the corner-marker shader block**; corner-marker removal is #380 (v2.1), out of scope here.

**[test] pin:** the producer -> `vtkMRMLLocatorNode` -> consumer chain delivers the world point to the mapper as `uLocatorPosition`.

Depends on Locator A; parallelizable with Locator B (the full chain pin closes when B+C are both in).

---
_Authored with assistance from Claude (Anthropic Claude Opus 4.8); part of the ADR-0025 locator, tracker #305._

RafaelPalomar:
Scope note: Slice C splits into two halves.

- **C++ shader half (in progress)** — `vtkOpenGLBezierResectionPolyDataMapper` gains `uLocatorPosition`/`uLocatorRadius` (Impl storage + Set/Get accessors + vertex RAS varying + fragment marker block that coexists with the corner-marker block per ADR-0025 Consequences, #380 removes the corner block later) + a GL-free accessor [test]. Branch `feature/locator-consumer` (stacked on Locator A / #487).
- **Python consumer Representation (deferred)** — wiring the T2 `LiverBezierSurfacePipeline` Representation to read `vtkMRMLLocatorNode` and push the uniforms is **blocked on the T2-mapper-relocation** (ADR-0014 §3): `BezierPlanningRepresentation` still uses a placeholder `vtkPolyDataMapper` (`TODO(T2-mapper-relocation)`), not the real mapper. The full node→consumer→`uLocatorPosition` chain closes once that relocation lands.

_Note from Claude (Opus 4.8) assisting the maintainer._

RafaelPalomar:
**Deferred test (tracked):** the interactive / visual-regression test of the locator marker *rendering* on the resection surface is owed here, with the consumer-wiring half — NOT with the shader-uniform half (#492).

[...]

PR #494, opened 2026-06-24
ENH: Wire BezierPlanningRepresentation to the real resection mapper (T2-mapper-wiring slice 1)

RafaelPalomar:
## T2-mapper-wiring slice 1 — drive `BezierPlanningRepresentation` off the real mapper

Closes the first slice of #493: the Planning-state LayerDM Representation now renders the Bézier resection surface through the relocated, real `vtkOpenGLBezierResectionPolyDataMapper` (ADR-0014 §3) instead of the placeholder raw control mesh through a generic `vtkPolyDataMapper`. This is the slice that **unblocks the ADR-0025 locator consumer (#489)**: a Python Representation can now call `SetLocatorPosition` / `SetLocatorRadius` (landed in #492) on a real, surface-rendering mapper instance.

### What changed
- **Surface mapper swap.** Inside a launched Slicer the Representation builds the tessellated Bernstein patch with `vtkBezierSurfaceSource` + `vtkPolyDataNormals` (resolution 20, carried verbatim from the v1 `vtkSlicerBezierSurfaceRepresentation3D` ctor) and drives the real `vtkOpenGLBezierResectionPolyDataMapper`. The source emits the 2-component `uvCoords` (TCoords) the mapper's vertex shader reads.
- **Display → uniform plumbing.** Ports `vtkSlicerBezierSurfaceRepresentation3D::UpdateBezierSurfaceDisplay`: resection / grid / margin colours, opacity, and (gated on `Grid3DVisibility`, v1 parity) grid divisions / thickness are pushed onto the mapper.
- **Bare-VTK fallback.** A `_resolve_vtk_class` helper selects the real path when the wrapped classes are reachable and falls back to a generic `vtkPolyDataMapper` over the raw control mesh in the unit layer, so the Representation still constructs and the colour bookkeeping stays testable.
- **Control-grid read fix.** Read the grid via `GetControlGridVector` (`std::vector<double>` → a Python tuple); VTK wraps the bare `const double*` of `GetControlGrid` as an opaque pointer-string, unusable from Python. Stub data nodes expose only `GetControlGrid`, so fall back to it.
- Retires the stale `TODO(T2-mapper-relocation)` tags (the relocation itself was already done; this is the wiring #493 re-scoped).

### Scoped out (deferred)
The RAS/IJK transform matrices and the distance-map 3D texture binding from #493 slice 1 are **not** wired here: the v2 `vtkMRMLBezierSurfaceNode` carries no distance-map reference (the v1 markups node had `GetDistanceMapVolumeNode()`), so there is no node-level source to thread. The fragment shader degrades gracefully when no `distanceTexture` is bound. Giving the v2 node a distance-map reference is a follow-on.

Per #493's open question, this slice **coexists** with the v1 markups render path (the cutover decision is deferred to a follow-on slice).

### Tests (invariant-test-first, ADR-0027)
New launched-Slicer pytest `test_bezier_planning_surface_mapper_wiring.py`, three GL-free invariants: the surface mapper is the real type; after `update()` the surface polydata is the tessellated patch (one point per resolution cell, with per-point normals and the `uvCoords` array) rather than the control mesh; and the display node's opacity / grid divisions / thickness reach the mapper's scalar uniform setters (the colour setters run on the same code path; their `const float*` getters are not readable from Python). The RED test was committed first and skips pending the swap; the implementation greens it. Skips cleanly under bare `PythonSlicer -m pytest`.

Verified locally under a launched Slicer: 56 passed / 9 skipped / 1 xfailed across `LiverResections/Testing/Python`, including the two new tests.

### Conformance
- ADR-0014 §3 — mapper relocation (done) + this wiring.
- ADR-0013 §5/§6 — LayerDM Pipeline / Representation; no custom DisplayableManager.
- ADR-0003 — offscreen-render abort: no distance texture is bound in this slice, and the shader's no-texture path is the documented fallback.
- ADR-0025 — the #489 locator consumer this unblocks.

---

### DIFF
--- a/LiverResections/LiverResectionsLib/Representations/BezierPlanningRepresentation.py
+++ b/LiverResections/LiverResectionsLib/Representations/BezierPlanningRepresentation.py
@@ -14,20 +14,26 @@ constructed by ``LiverBezierSurfacePipeline.initialize()`` and
 exercised by direct Python instantiation in unit tests.  Full
 integration with Slicer's LayerDM framework happens at T2.6.
 
-Per ADR-0014 §3 the four custom OpenGL mappers
-(``vtkOpenGLBezierResectionPolyDataMapper``,
-``vtkOpenGLSlicingContourPolyDataMapper``,
-``vtkOpenGLDistanceContourPolyDataMapper``,
-``vtkOpenGLResection2DPolyDataMapper``) **relocate** from
-``LiverMarkups/VTKWidgets/`` to ``LiverResections/VTKWidgets/``.  The
-relocation has not landed yet (it is part of the broader T2 cohort,
-not the T2.2 stack); this Representation uses the generic
-``vtkPolyDataMapper`` + ``vtkActor`` pair until that relocation
-completes, at which point the surface mapper field will flip to
-``vtkOpenGLBezierResectionPolyDataMapper`` *without changing the
-Representation's public API*.  Marked with
-``TODO(T2-mapper-relocation)`` at the construction point so the
-swap is mechanical.
+Surface mapper (T2-mapper-wiring)
+---------------------------------
+Per ADR-0014 §3 the four custom OpenGL mappers relocated from
+``LiverMarkups/VTKWidgets/`` to ``LiverResections/VTKWidgets/``.  Inside a
+launched Slicer this Representation drives the relocated, real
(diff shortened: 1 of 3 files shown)


RafaelPalomar (commented):
LGTM

PR #496, opened 2026-06-25
ENH: Distance-map input on the resection-plan wrapper + bind it on the v2 render path (ADR-0031)

RafaelPalomar:
## Distance-map input on the resection-plan wrapper + bind it on the v2 render path

Implements ADR-0031 and closes the prerequisite for the #493 render cutover: the v2 LayerDM Bézier render path now reproduces v1's distance-map margin / uncertainty / clip-out shading, so v1's live Bézier render can be removed without a visual-regression.

Two slices, test-first (ADR-0027):

### 1 — `distanceMap` node reference on `vtkMRMLResectionPlanNode` (commit 1)
The distance-map volume is a path-specific input of the resection plan, so per ADR-0031 it lives on the **wrapper** (typed `distanceMap` ref role: `GetDistanceMapReferenceRole()` / `GetDistanceMapVolumeNode()` / `SetAndObserveDistanceMapVolumeNode()`), alongside the `SafetyMargin_mm` / `RiskMargin_mm` scalars already there — not on the carrier. Mirrors the existing `geometry` role. MRML-level invariant test pins the role literal, the in-scene round-trip, and nullptr-clears.

### 2 — bind the distance map on the render path (commit 2)
- **C++ (mapper):** `SetDistanceMapImageData(vtkImageData*)` stores the image and builds + binds the 3D distance texture lazily in `BuildBufferObjects` on a live GL context, rebuilding only when the image MTime advances (realize-then-bind — ADR-0003). The `void*` `Create3DFromRaw` upload can't cross the Python wrap, so it lives in C++. A single base `vtkTextureObject` suffices — no LiverMarkups multi-texture-helper dependency. `GetDistanceMapImageData()` is the GL-free introspection point.
- **Python (Representation):** `SetResectionPlanNode()`; `update()` reads the wrapper's distance-map volume + margins, hands the image to the mapper, computes the RAS→IJK / IJK→texture matrices from the volume geometry, and pushes the margins. No distance map → clears the image + resets matrices to identity (graceful fallback). The Pipeline threads the wrapper to the Representation before `update()`.

### Tests
- C++ GL-free accessor test (`vtkMRMLResectionPlanDistanceMapRefTest`, `vtkOpenGLBezierResectionPolyDataMapperDistanceMapTest1`).
- Launched-Slicer invariant (`test_planning_distance_map_threaded_from_plan_to_mapper`): the full plan→Representation→mapper thread (image stored, margins set, RAS/IJK matrix non-identity, clearing returns to the fallback).
- Texture build + bind + a two-render **no-ADR-0003-abort** check verified on an interactive `:0` render.

Verified locally: all MRML + VTKWidgets C++ tests pass; the launched LiverResections suite is 57 passed / 9 skipped / 1 xfailed (no failures introduced).

### Scope / conformance
This is the binding prerequisite only. Removing the v1 `vtkSlicerBezierSurfaceWidget` / `vtkSlicerBezierSurfaceRepresentation3D` render and re-baselining the `BezierSurface4x4*` visual-regression scenarios are the follow-on cutover slices tracked under #493. References: ADR-0031, ADR-0014 §"Fourth layer", ADR-0013 §4/§5, ADR-0003, ADR-0027.

---

### 3 — Pipeline reverse-resolves the plan wrapper (commit 3)
The LayerDM displayable manager creates the Pipeline for a surface's `vtkMRMLParametricSurfaceDisplayNode`, so it derived the data node but was never handed the wrapper — nothing in production calls `SetResectionNode`, so the live v2 surface rendered with **no margin shading**. `UpdatePipeline` now reverse-resolves the `vtkMRMLResectionPlanNode` from the data node's `geometry` back-reference and adopts it, so the distance map + margins reach the mapper through the slice-2 path with no external caller. Launched invariant test pins the resolution + the negative (unowned-surface → no wrapper) case. This completes the **live** v2 distance-map render path; without it slices 1+2 have no production driver.

### DIFF
--- a/LiverResections/LiverResectionsLib/LiverBezierSurfacePipeline.py
+++ b/LiverResections/LiverResectionsLib/LiverBezierSurfacePipeline.py
@@ -287,6 +287,19 @@ class LiverBezierSurfacePipeline(_PipelineBase):
         """
         self._ensure_representations()
 
+        # Reverse-resolve the orchestrating wrapper when no explicit one was
+        # set (ADR-0031).  The LayerDM creator hands the Pipeline only the
+        # display node (it matches on vtkMRMLParametricSurfaceDisplayNode), so
+        # nothing sets _resection_node in the live render path; discover the
+        # vtkMRMLResectionPlanNode whose geometry reference is our data node
+        # and adopt it, so the plan's distance-map + margins reach the mapper
+        # with no external SetResectionNode caller.  Re-attempted while unset
+        # (the plan's geometry ref may be wired after the display node lands).
+        if self._resection_node is None and self._data_node is not None:
+            resolved = self._resolve_resection_node()
+            if resolved is not None:
+                self.SetResectionNode(resolved)
+
         # Build the dispatch key.  Falls back to ``0`` MTime for nodes
         # whose ``GetMTime`` is unavailable (defensive — stub nodes in
         # tests may not implement it).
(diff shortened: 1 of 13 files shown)


RafaelPalomar (commented):
LGTM

PR #497, opened 2026-06-25
ENH: Python-wrappable per-point control-grid setter on the parametric surface

RafaelPalomar:
## Python-wrappable per-point control-grid setter on the parametric surface

A small node-API enabler for the render-cutover re-baseline (#493).

### Why
`vtkMRMLAbstractParametricSurfaceNode::SetControlGrid` takes a bare `const double*`, which VTK cannot wrap for Python (it surfaces as an opaque pointer string — the same wall as `GetControlGrid`). So a Python scenario or interaction code has **no way to position the v2 surface's control grid**. That blocks the re-baseline: every existing visual-regression scenario still hand-builds the **v1 markups** node because no v2 grid setter is reachable from Python.

### What
`SetControlPoint(row, col, x, y, z)` — scalar arguments cross the Python wrap cleanly. Writes the `(row, col)` triple into the row-major flat grid at `(row * Cols + col) * 3`; returns `false` (grid untouched) for out-of-range `row`/`col`. Declared on the abstract base so the v2.1 NURBS sibling shares the seam.

### Tests
- C++ invariant (`vtkMRMLAbstractParametricSurfaceNodeTest1`): round-trips a control point through the flat grid + out-of-range rejection, via the abstract-base pointer.
- Launched-Slicer test (`test_parametric_surface_setcontrolpoint.py`): pins that `SetControlPoint` is callable **from Python** and round-trips via `GetControlGridVector` — the actual contract this slice provides.

Both pass locally; the Python test skips cleanly under bare `PythonSlicer -m pytest`.

### Context
Independent of #496 (the distance-map render path); together they unblock porting the `BezierSurface4x4*` visual-regression scenarios to drive the v2 surface for re-baselining. References: ADR-0018 §1 (grid-size invariants), ADR-0014 §"Fourth layer", #493 (the cutover).

---

### DIFF
--- a/LiverResections/MRML/Testing/Cxx/vtkMRMLAbstractParametricSurfaceNodeTest1.cxx
+++ b/LiverResections/MRML/Testing/Cxx/vtkMRMLAbstractParametricSurfaceNodeTest1.cxx
@@ -447,6 +447,38 @@ int testGetInitModeAsStringInvalidDefault()
   return EXIT_SUCCESS;
 }
 
+//------------------------------------------------------------------------------
+// Per-point control-grid setter -- the Python-wrappable seam a scenario /
+// interaction uses to position the grid (the flat ``SetControlGrid(const
+// double*)`` cannot cross the Python wrap).  ``SetControlPoint(row, col, x,
+// y, z)`` writes the (row, col) triple into the row-major flat grid at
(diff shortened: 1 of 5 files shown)


RafaelPalomar (commented):
LGTM

Issue #490, opened 2026-06-21
Locator D: click-to-reslice wiring


RafaelPalomar:
Clicking in the resectogram updates the orthogonal slice's `SliceToRAS` so it passes through the picked world point; the consumer side observes `vtkMRMLLocatorNode` and drives the slice.

**[test] pin:** click-to-reslice updates the slice `SliceToRAS` so the slice passes through the picked world point.

Depends on Locator A; follows Locator B.

---
_Authored with assistance from Claude (Anthropic Claude Opus 4.8); part of the ADR-0025 locator, tracker #305._

RafaelPalomar:
Audit against merged `preview`: shipped and exceeded.

`LocatorReslicer` observes the scene's `vtkMRMLLocatorNode` and reslices via `vtkMRMLSliceNode.JumpSliceByOffsetting` — orientation-preserving, the slice plane passes through the picked world point (PR #524). Extended to jump **all** slice views rather than the orthogonal one only (PR #564). Pinned by `test_locator_reslice_consumer.py` (including the jump-all pin with independently oriented slice nodes). Closing.

Issue #488, opened 2026-06-21
Locator B: T3 resectogram producer Representation (pixel -> (u,v) -> world)


RafaelPalomar:
Producer Representation in the T3 `ResectogramPipeline`: map a resectogram pixel via `vtkLiverResectogramPixelMapping::PixelToUV` then `vtkMRMLBezierSurfaceNode::EvaluateSurface(u,v)` to an **exact** world point, and write it onto `vtkMRMLLocatorNode`. **No `vtkCellPicker`** (ADR-0025 Alternative A rejected — the resectogram IS the (u,v) image, so the map is exact).

**[test] pin:** Bézier surface evaluation at a known (u,v) matches the expected world-space point (pure, no render context).

Depends on Locator A. **Open question:** how the producer receives click pixel coordinates from the embedded resectogram widget (Qt observer vs VTK interactor callback) — tied to #483.

---
_Authored with assistance from Claude (Anthropic Claude Opus 4.8); part of the ADR-0025 locator, tracker #305._

RafaelPalomar:
Audit against merged `preview`: shipped.

`ResectogramLocatorProducer` maps pixel → `vtkLiverResectogramPixelMapping.PixelToUV` → `vtkMRMLBezierSurfaceNode.EvaluateSurface(u, v)` → exact world point onto `vtkMRMLLocatorNode` — no `vtkCellPicker`, per ADR-0025's rejection of Alternative A (PR #516; the row-first/column-first transpose fix rode PR #559). Pinned by `test_resectogram_locator_producer.py`. Closing.

PR #515, opened 2026-07-07
ENH: Locator consumer — drive the resection shader marker off vtkMRMLLocatorNode (ADR-0025)

RafaelPalomar:
Contributes to #489 (consumer half; the shader half #492 is already merged).

## Summary

Implements the ADR-0025 **locator consumer** in the T2 `LiverBezierSurfacePipeline`: the scene's single `vtkMRMLLocatorNode` picked point drives the resection-surface shader's locator marker. No new displayable manager / pipeline factory (ADR-0013 §5) — a consumer Representation inside the existing pipeline.

## What lands

- **`BezierPlanningRepresentation.SetLocatorNode(node)`** + `_apply_locator()`: reads `GetPickedPositionWorld()` off the locator node and `Radius` off its display node, driving the real `vtkOpenGLBezierResectionPolyDataMapper`'s `SetLocatorPosition` / `SetLocatorRadius` (the `uLocatorPosition` / `uLocatorRadius` uniforms from #492). No locator node → radius `0` (marker off). Guarded (getattr) so the bare-VTK unit layer's generic mapper is tolerated.
- **`LiverBezierSurfacePipeline`**: reverse-resolves the scene's single `vtkMRMLLocatorNode` (scene-MTime-gated, mirroring the resection-plan resolve), observes it so picked-point changes re-dispatch, folds its MTime into the memo-key (so off→on re-dispatches), and threads it onto the active Representation's `SetLocatorNode` seam before `update`.

## Tests (test-first, ADR-0027)

- New `test_bezier_planning_locator_consumer.py` + `test_pipeline_resolves_locator.py`, RED-as-skip on the seam, green after.
- Launched suite: 93 passed / 14 skipped / 1 xfailed / 0 failed. Bare-VTK unit layer: 84 passed (the guarded `_apply_locator` leaves it green).
- The **2 pipeline-resolve assertions skip headless** ("no active real-mapper Representation without a realized view") — the same gate the other pipeline-routing tests use. The rep-level consumer (node → mapper uniform) is directly proven; the resolve/thread code runs clean through the full suite. The pipeline-resolve + actual marker render is an interactive `:0` follow-up, and the full **producer → node → consumer** chain closes with Slice B (#488).

## Scope
Consumer only. Producer (#488, resectogram click → (u,v) → world → node) and click-to-reslice (#490) follow.

### DIFF
--- a/LiverResections/LiverResectionsLib/LiverBezierSurfacePipeline.py
+++ b/LiverResections/LiverResectionsLib/LiverBezierSurfacePipeline.py
@@ -197,6 +197,15 @@ class LiverBezierSurfacePipeline(_PipelineBase):
         # dispatch (reset when the display/data node changes).
         self._last_resection_scan_mtime: int | None = None
 
+        # The cross-view locator node (ADR-0025), reverse-resolved from the
+        # scene (v2.0 has exactly one).  Observed so its picked-point changes
+        # re-dispatch; threaded onto the active Representation's consumer seam.
+        # Discovery is gated on the scene ``MTime`` (like the plan scan) so a
+        # position update to an already-resolved locator rides its own observer
+        # rather than a per-tick class scan.
+        self._locator_node: Any | None = None
+        self._last_locator_scan_mtime: int | None = None
+
         # Observer tags so ``cleanup()`` can detach precisely.  Index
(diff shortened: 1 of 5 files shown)


RafaelPalomar (commented):
LGTM

PR #518, opened 2026-07-07
BUG: Init-mode ring extraction silent no-op — resolve extractors from Algorithm wrapping

### LINKED ISSUE #517: BUG: Init-mode ring extraction is a silent no-op — extractor classes resolved from wrong namespace
## Summary

The SlicingPlane and DistanceSpheroid init-mode Representations resolve their
ring-extractor classes from the wrong Python namespace, so ring extraction is
a **permanent silent no-op** in launched Slicer — the intersection-ring
overlays are never computed.

## Root cause

`_resolve_extractor_class` (in both `SlicingPlaneInitRepresentation.py` and
`DistanceSpheroidInitRepresentation.py`) looks up the wrapped-C++ class over
`("slicer", "vtk")`:

```python
for module_name in ("slicer", "vtk"):
    module = __import__(module_name)
    cls = getattr(module, name, None)
    ...
return None
```

But `vtkLiverPlaneRingExtractor` and `vtkLiverSpheroidRingExtractor` are
**Algorithm-library** classes. A launched-Slicer namespace probe shows they
are exposed **only** on `vtkSlicerLiverResectionsModuleAlgorithmPython` —
not on `slicer`, not on `vtk`:

```
vtkLiverPlaneRingExtractor      -> ['vtkSlicerLiverResectionsModuleAlgorithmPython']
vtkLiverSpheroidRingExtractor   -> ['vtkSlicerLiverResectionsModuleAlgorithmPython']
```

So the resolver returns `None` on every call, and `run_ring_extraction`
bails without ever computing `_ring_polydata`. The unit tests pass because
they assert the graceful-`None` degradation path; the interactive pass would
surface the missing rings.

This is the same namespace mismatch that affects the Algorithm-library
`vtkLiverResectogramPixelMapping`, which was caught before shipping in the
locator producer.

## Secondary cleanup

The same probe confirms **no** Liver class is reachable from the plain
upstream `vtk` module (VTKWidgets classes are on `slicer`, Algorithm classes
on `AlgorithmPython`). The `vtk` fallback in all four `("slicer", "vtk")`
resolvers (`_resolve_extractor_class` x2, `_require_vtk_class` x2) is dead —
the "fallback for non-Slicer VTK builds" comment is inaccurate. Each resolver
should collapse to a lazy import of the single namespace that actually holds
its class.

## Fix (test-first, ADR-0027)

1. RED test: assert `run_ring_extraction` returns a **non-None** ring polydata
   for a known plane / spheroid under launched Slicer (fails today).
2. Point the extractor resolvers at `vtkSlicerLiverResectionsModuleAlgorithmPython`.
3. Simplify all four resolvers to a single-namespace lazy import; drop the
   dead `vtk` fallback.

## Verified namespace map (launched Slicer)

```
vtkLiverResectogramPixelMapping         -> AlgorithmPython
vtkLiverPlaneRingExtractor              -> AlgorithmPython
vtkLiverSpheroidRingExtractor           -> AlgorithmPython
vtkOpenGLBezierResectionPolyDataMapper  -> slicer + VTKWidgetsPython
vtkOpenGLDistanceContourPolyDataMapper  -> slicer + VTKWidgetsPython
vtkBezierSurfaceSource                  -> slicer
```

Relates to ADR-0014 §3 (wrapped-class relocation) and ADR-0025 (locator).

RafaelPalomar:
## Summary

Fixes a **silent no-op** in the SlicingPlane and DistanceSpheroid init-mode
ring extraction. `_resolve_extractor_class` looked up
`vtkLiverPlaneRingExtractor` / `vtkLiverSpheroidRingExtractor` over the
`(slicer, vtk)` namespaces, but those are **Algorithm-library** classes
(ADR-0014 §3) exposed inside a launched Slicer **only** on
`vtkSlicerLiverResectionsModuleAlgorithmPython`. The resolver returned `None`
on every call, so `run_ring_extraction` never computed an intersection ring —
the overlays were permanently absent, and the unit tests passed only because
they asserted the graceful-`None` degradation path.

This is the same namespace mismatch caught in the locator producer
(`vtkLiverResectogramPixelMapping` is Algorithm-library too).

Fixes #517.

## Fix

Point the extractor resolvers at the Algorithm Python wrapping, lazily
imported so the module stays importable where the wrapping is off the path
(the bare-VTK unit layer).

## Simplification

The same launched-Slicer namespace probe confirmed **no** Liver class is
reachable from the plain upstream `vtk` module (VTKWidgets classes are on
`slicer`, Algorithm classes on `AlgorithmPython`). All three resolvers
therefore collapse from the `for module_name in ("slicer", "vtk")` +
`__import__` loop to a single-namespace lazy import — the `vtk` fallback was
dead code. The VTKWidgets mapper/source resolvers (`_require_vtk_class`) keep
their resolve-or-raise contract, now via `slicer` only (−77/+62 lines).

Verified namespace map (launched Slicer):

```
vtkLiverPlaneRingExtractor              -> AlgorithmPython
vtkLiverSpheroidRingExtractor           -> AlgorithmPython
vtkOpenGLBezierResectionPolyDataMapper  -> slicer + VTKWidgetsPython
vtkOpenGLDistanceContourPolyDataMapper  -> slicer + VTKWidgetsPython
vtkBezierSurfaceSource                  -> slicer
```

## Test-first (ADR-0027)

Landed RED-then-green. The new invariant test asserts the extractor classes
resolve to a real class (not `None`) and that a plane through a sphere centre
yields a non-empty ring — it **fails** under the launched-Slicer row before
the fix and passes after. It skips cleanly under bare `pytest` (the Algorithm
wrapping is off that path).

Verified launched: 14/14 pass in the ring-extraction + mapper-wiring set
(1 pre-existing unrelated skip); the three new assertions flipped RED→GREEN,
and the Bezier / DistanceSpheroid / Vascular / Flattened mapper-wiring tests
confirm the `_require_vtk_class` change did not regress VTKWidgets resolution.

---

### DIFF
--- a/LiverResections/LiverResectionsLib/Representations/BezierPlanningRepresentation.py
+++ b/LiverResections/LiverResectionsLib/Representations/BezierPlanningRepresentation.py
@@ -726,31 +726,28 @@ def _identity_matrix() -> Any:
 
 
 def _require_vtk_class(name: str) -> Any:
-    """Resolve a wrapped-C++ class by name from ``slicer`` then ``vtk``, or raise.
+    """Resolve a wrapped-C++ VTKWidgets class by name from ``slicer``, or raise.
 
     The relocated mapper + the Bezier surface source are wrapped into Slicer's
-    ``slicer`` namespace (``SlicerMacroBuildModuleLogic`` Python wrapping); the
-    plain ``vtk`` module is the fallback for a non-Slicer VTK build.  Raises
-    ``RuntimeError`` when neither namespace exposes the class — a real
-    misconfiguration in production (ADR-0014 §3) must fail loudly rather than
+    ``slicer`` namespace (ADR-0014 §3).  Raises ``RuntimeError`` when a class is
+    absent — a real misconfiguration in production must fail loudly rather than
     degrade to a shader-less generic mapper.  Bare-VTK unit tests (ADR-0008 §2)
     avoid this path by injecting a mapper instance instead.
     """
-    for module_name in ("slicer", "vtk"):
-        try:
-            module = __import__(module_name)
-        except ImportError:
-            continue
-        cls = getattr(module, name, None)
-        if cls is not None:
-            return cls
-    raise RuntimeError(
-        f"{name} is not reachable from the 'slicer' or 'vtk' namespace. "
-        "It is a wrapped-C++ class relocated to LiverResections/VTKWidgets/ "
-        "(ADR-0014 §3) and available only inside a launched Slicer with the "
-        "module loaded.  Inject a mapper instance for bare-VTK unit tests "
-        "(ADR-0008 §2)."
-    )
+    try:
+        import slicer
+    except ImportError:
+        slicer = None
+    cls = getattr(slicer, name, None) if slicer is not None else None
+    if cls is None:
+        raise RuntimeError(
+            f"{name} is not reachable from the 'slicer' namespace. "
+            "It is a wrapped-C++ class relocated to LiverResections/VTKWidgets/ "
+            "(ADR-0014 §3) and available only inside a launched Slicer with the "
+            "module loaded.  Inject a mapper instance for bare-VTK unit tests "
+            "(ADR-0008 §2)."
+        )
+    return cls
 
 
 def _push_color3(mapper: Any, setter_name: str, getter: Any | None) -> None:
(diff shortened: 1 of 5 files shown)


RafaelPalomar (commented):
LGTM

PR #519, opened 2026-07-07
ENH: Resectogram locator interaction seam on the Pipeline — ADR-0032 / ADR-0025

RafaelPalomar:
## Summary

Wires the **resectogram locator interaction seam** (ADR-0032; ADR-0025
§Producer / §Click-to-reslice) — the last piece of the cross-view locator
stack. A left click on the embedded resectogram view now sources the locator
pick that the already-merged consumer renders as a marker on the 3D Bézier
surface.

The seam lives entirely on `ResectogramPipeline` (no widget↔pipeline coupling,
ADR-0013 §5 clean):

- `_produce_from_display_position(display_xy)` — the GL-free core. Resolves the
  four producer inputs off the live Pipeline state — surface carrier
  (`GetDataNode`), the scene's single `vtkMRMLLocatorNode`, the `mat_ratio`
  (`GetFlattenedSurfaceRepresentation().GetMatRatioApplied()`) and the strip
  renderer's render-window size — then composes
  `ResectogramLocatorProducer(surface, locator).produce(...)`, the exact 1:1
  (u, v) mapping (no `vtkCellPicker`). Each degenerate input (no surface / no
  locator / unresolved MatRatio / non-positive viewport) is a no-op returning
  `None`, so a half-configured pipeline never feeds a false pick to the
  consumer.
- `CanProcessInteractionEvent` claims resectogram-view interaction whenever a
  surface is displayed (the strip *is* the (u, v) domain; every in-view pixel
  maps). `ProcessInteractionEvent` reads `eventData.GetDisplayPosition()` (VTK
  bottom-left origin — no Qt y-flip) and drives the seam, mirroring
  `LiverBezierSurfacePipeline`.

Contributes to #488.

## Test-first (ADR-0027)

Landed RED-then-green. The new invariant reuses the sibling producer test's
known affine 4×4 grid + composition anchor: VTK display pixel (64, 192) in a
256×256 viewport at isotropic ratio → (u, v) = (0.25, 0.75) → world
(22.5, 7.5, 0), written onto the locator's `PickedPositionWorld`; plus four
degenerate no-op guards. GL-free — the renderer and render window are faked.

Verified launched: **5/5 pass, 0 skipped** (the composition test runs, not
skips — no green-but-skipping). Skips cleanly under bare `pytest` (LayerDMLib /
wrapped classes off that path).

## Scope / remaining verification

GL-free seam only. Whether `ProcessInteractionEvent` fires past the embedded
view's camera left-button lock (`lockEmbeddedViewInteraction`) is the risk
ADR-0032 §Consequences pre-registered — resolvable only on an interactive
display. If the event is swallowed by the lock, the sanctioned fallback is a Qt
event filter on the view (Option A) plus an ADR-0032 amendment.

---

### DIFF
--- a/LiverResections/LiverResectionsLib/ResectogramPipeline.py
+++ b/LiverResections/LiverResectionsLib/ResectogramPipeline.py
@@ -67,6 +67,7 @@ References
 .. _ADR-0013: ../../Docs/adr/0013-layerdm-pipeline-pattern.md
 .. _ADR-0015: ../../Docs/adr/0015-cpp-algorithm-library.md
 .. _ADR-0025: ../../Docs/adr/0025-locator-architecture.md
+.. _ADR-0032: ../../Docs/adr/0032-v2-interaction-via-layerdm-pipeline-seam.md
 """
 
 from __future__ import annotations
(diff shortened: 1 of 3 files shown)


RafaelPalomar (commented):
LGTM

PR #524, opened 2026-07-07
ENH: Click-to-reslice consumer (LocatorReslicer) — ADR-0025 Slice D

RafaelPalomar:
## Summary

Implements the ADR-0025 **click-to-reslice** consumer (§Click-to-reslice) — the
last data-flow arrow of the cross-view locator, `Node → Slice`, distinct from
the shader-marker path `Node → Consumer → Shader` that already merged.

`LocatorReslicer` is a plain Python observer (ADR-0013 §5 clean — not a Pipeline
or displayable manager; ADR-0004 Python): it resolves the scene's single
`vtkMRMLLocatorNode`, observes its `ModifiedEvent`, and on a picked-point change
reslices the Red slice node so its plane passes through the point.

The pure kernel `reslice_slice_to_world(slice_node, world_xyz)` uses
`vtkMRMLSliceNode.JumpSliceByOffsetting` — a translation along the slice normal
only, so orientation is preserved and only the offset moves onto the point. It
reads the picked point off the data-only locator carrier (ADR-0014) and writes
only the slice node.

Drives the **Red** slice for v2.0; a Red/Yellow/Green sweep is a v2.1
enhancement (alongside the deferred cross-module locator unification, ADR-0012).

Contributes to #490.

## Test-first (ADR-0027)

Landed RED-then-green. The [test] pin (verbatim from #490): *"click-to-reslice
updates the slice SliceToRAS so the slice passes through the picked world
point."* The kernel test seeds a known non-axial `SliceToRAS`, picks a point off
the plane, reslices, and asserts (a) the point now lies in the plane
(`dot(normal, point − origin) ≈ 0`, read off the matrix) and (b) the normal is
unchanged (orientation preserved); plus degenerate no-ops.

Verified launched: **3/3 kernel tests pass**. GL-free — `SliceToRAS` lives on
the bare slice node, no view/widget realised; skips cleanly under bare pytest. A
secondary scene-observer test rides the launched Red slice node and skips
headless (`--no-main-window` has no default slice nodes).

## Scope

Tested consumer unit only. The live wiring — creating the `vtkMRMLLocatorNode`,
having `ResectionPlanningWidget` own the `LocatorReslicer`, and `:0`-verifying
click→reslice + click→marker end-to-end — is a separate integration step (the
locator chain is currently inert because nothing instantiates the node yet).

---

### DIFF
--- a/LiverResections/LiverResectionsLib/LocatorReslicer.py
+++ b/LiverResections/LiverResectionsLib/LocatorReslicer.py
@@ -1,0 +1,122 @@
+# Copyright (c) 2026, The Intervention Centre, Oslo University Hospital. All rights reserved.
+# Distributed under the OSI-approved BSD 3-Clause License.
+"""Click-to-reslice consumer — locator picked point drives the slice (ADR-0025).
+
+The cross-view locator's picked RAS world point (carried on the single
+``vtkMRMLLocatorNode``, `ADR-0025`_ §"The node") drives the orthogonal slice so
+its plane passes through that point ("click-to-reslice", `ADR-0025`_
+§Click-to-reslice).  This is the data-flow arrow ``Node -> Slice``, distinct
+from the shader-marker arrow ``Node -> Consumer -> Shader`` the T2 Pipeline
+already carries.
+
+Per `ADR-0013`_ §5 this is **not** a Pipeline or a displayable manager — it is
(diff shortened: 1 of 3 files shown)


RafaelPalomar (commented):
LGTM

Issue #527, opened 2026-07-07
Stage 2 (Anatomy) usable end-to-end without in-app AI: load/tag canonical segmentation + guard silent hand-off + fix completion routing


RafaelPalomar:
## Problem

Stage 2 (Anatomy Definition / LiverSegmentation) cannot produce a usable
canonical segmentation for the end-to-end v2.0 flow:

1. **AI segmentation is stubbed.** `LiverSegmentation.py:509` `_runTotalSegmentator`
   calls the install-wrapper's `run()` but returns an *empty* scratch node
   (`:513` `createScratchSegmentation()`); the comment states real inference
   "lands with the backend integration." So "Run TotalSegmentator" produces no
   segments. Per the v2.0 scope decision, **in-app AI inference is deferred to
   v2.1** (tracked separately) — v2.0 gets its canonical segmentation by
   **loading/importing an existing segmentation and tagging it canonical**.
2. **Silent hand-off failure.** `selectInputVolume` (`LiverSegmentation.py:481`)
   returns `None` when no volume is tagged `LiverRole='PortalVenous'`, and
   `_StructureCard.onRun` (`:321`) then calls `segment(None)` unguarded (`:322`,
   `:325`) — no user feedback; the workflow appears to run but does nothing.
3. **Completion predicate mis-routed.** `Liver.py` routes Stage-2 completion to
   the always-False stub `_stage2IsComplete` (`:472`) instead of the module's
   own `isStageComplete`, so the sidebar never shows Stage 2 done.

## Scope (v2.0)

- Support obtaining a canonical `vtkMRMLSegmentationNode` **without in-app AI**:
  load/import a segmentation and tag it canonical (the `LiverSegmentation.Role='canonical'`
  attribute the downstream stages read), and/or the manual Segment Editor path.
- Guard the `selectInputVolume() -> None` path: disable Run + surface "tag a
  PortalVenous volume in Case Setup first" rather than calling `segment(None)`.
- Route Stage-2 completion to the module's `isStageComplete` when the module is
  present (`Liver.py` stage-completion dispatch).

## Notes
End-to-end blocker for a fully-usable v2.0. Depends on #523 (Case Setup tags the
`LiverRole='PortalVenous'` volume this stage selects). In-app TotalSegmentator
inference is a separate v2.1 issue.

RafaelPalomar:
Audit against merged `preview`: all three scope items shipped.

[...]

PR #531, opened 2026-07-07
ENH: Stage-2 completion routing + silent hand-off guard (#527 Slice A)

RafaelPalomar:
## Summary

#527 **Slice A** — two pure-Python fixes making the Stage-2 (Anatomy) workflow
honest end-to-end. First of three slices toward Stage 2 being usable without
in-app AI (Slice B = Stage-3 SCT liver lookup; Slice C = load-a-segmentation +
per-structure SCT tagging). Contributes to #527.

## Fixes

- **Stage-2 completion routing** (`Liver.py:_stageIsComplete`). The
  `shellPredicate` dict pinned row 1 to the always-False `_stage2IsComplete`
  stub, short-circuiting before the module-logic lookup — so a completed anatomy
  stage never flipped its indicator. Dropped the row-1 entry. Because
  LiverSegmentation is a **scripted** module whose `isStageComplete` lives on the
  Python `LiverSegmentationLogic` (not the wrapped `module.logic()`), the module
  path now resolves a fresh `LiverSegmentationLogic` instance — it reads the same
  scene, so no module widget is built — and falls back to the graceful-
  degradation `False` when the module is absent (ADR-0023 §Stage 2).
- **Silent hand-off guard** (`LiverSegmentation.py:_StructureCard.onRun`). When
  `selectInputVolume()` is `None` (no `LiverRole='PortalVenous'` volume tagged in
  Case Setup), `onRun` called `segment(None)` unguarded — a silent no-op that
  read as success. Now it surfaces "Tag a PortalVenous volume in Case Setup
  (Stage 1) first" and leaves Accept/Reject disabled (ADR-0024 hand-off).

## Test-first (ADR-0027)

Extends two existing suites (no new files). Routing: `_stageIsComplete(1)` is
True with a canonical SCT-tagged segmentation, False on an empty scene, and
rows 0/5 + the injection override don't regress — via a setup-free
`_liver_widget_no_setup` so it's verifiable headless. Guard: a card Run with no
PortalVenous volume produces no scratch node + surfaces feedback; a tagged
volume still segments.

**6/6 pass launched** (routing + guard). GL-free; skip cleanly bare / when the
module is absent.

---

### DIFF
--- a/Liver/Liver.py
+++ b/Liver/Liver.py
@@ -627,11 +627,17 @@ class LiverWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
     shellPredicate = {
       0: self._stage1IsComplete,
-      1: self._stage2IsComplete,  # Stage 2 stub — used when LiverSegmentation is absent.
       5: self._stage6IsComplete,
     }.get(row)
     if shellPredicate is not None:
       return shellPredicate()
+
+    # Stage 2 (row 1) routes to the LiverSegmentation module's own
+    # isStageComplete via the module-logic lookup below when the module is
+    # present; when it is absent the lookup returns False (the graceful-
+    # degradation semantics _stage2IsComplete documents).  It is NOT pinned to
+    # the always-False shell stub here, or a completed anatomy stage would never
+    # flip its indicator (ADR-0023 §Stage 2).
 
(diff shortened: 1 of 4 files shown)


RafaelPalomar (commented):
LGTM

Issue #538, opened 2026-07-08
Wire distance-map computation into the Stage-2 (Anatomy) accept transition


RafaelPalomar:
## Problem

The v2 stage workflow **never computes a distance map**. A real-anatomy walkthrough (load volume + segmentation, drive Case Setup → Anatomy → Planning) reaches Stage 4 with **no distance-map volume in the scene**. The distance map is *consumed* throughout — `vtkMRMLResectionPlanNode`'s distance-map reference, `ResectogramPipeline` / `FlattenedSurfaceRepresentation`, and the OpenGL resection mappers (`vtkOpenGLBezierResectionPolyDataMapper`) — but no stage produces it.

Consequences:
- The resectogram strip cannot texture (`Failed to determine texture parameters`); it renders empty/white with only the colour border.
- Distance-based resection feedback has no input.

## Decision

Compute the distance map(s) on **Stage 2 (Anatomy) accept**, from the tagged liver segment (+ targets), producing the distance-map volume the resection plan references. Pairs the compute with the Stage-2 accept/contract action (see phase-contracts work) so Stage 4 opens with the map ready.

## Scope / open questions

- Confirm whether a distance-map computation already exists to wire (v1 path) vs. needs to be built in v2.
- Progress/feedback surface during compute (can be slow on full volumes).

## Refs

ADR-0023 §Stage 2 / §Stage 4; ADR-0025 §Context (resectogram is the flattened distance-map image). Related: #537 (representation lifecycle), #440 (phase contracts — the Stage-2 accept trigger).

RafaelPalomar:
Audit against merged `preview`: shipped, with one variant from the Decision.

`ensureDistanceMap` computes the distance map on **canonical import** rather than on the Stage-2 accept transition (PR #553) — the accept-contract action this issue proposed to pair with stayed in v2.1 (#440), and the import trigger achieves the stated outcome: Stage 4 opens with the tagged map ready, `CreateResectionPlan` auto-attaches it, and the resectogram textures (no `Failed to determine texture parameters`). Pinned by `test_liversegmentation_distance_map_on_import.py` and verified in repeated real-anatomy demo sessions.

If the trigger should later move onto the accept transition, that lands with #440's contract work in v2.1 — the compute + attach machinery is in place either way. Closing.

Issue #440, opened 2026-06-08
v2.1: User-verified phase contracts + soft-gating for the Liver shell


RafaelPalomar:
## Summary

Promote the Liver shell's per-stage state indicators from **advisory**
(v2.0) to **user-verified phase contracts with soft-gating** (v2.1).

Each stage exposes a *contract* the surgeon explicitly **verifies** (a
human attestation, not an automated clinical threshold). Verification
**unlocks the relevant downstream stages** along the ADR-0023
§"Cross-stage dependencies" map, and the stage's tab gains a pass
indicator. Not all stages are mandatory, so the unlock graph is a DAG
with skip paths, not a linear chain.

## Background

- v2.0 (T5.2-d, PR #438) shipped the shell as a vertical `QTabWidget`
  with **advisory, non-blocking** `✓ / ● / ○` indicators computed from
  each stage's `isStageComplete()` predicate. That is the
  forward-compatible substrate this issue builds on — no rework needed.
- The design is captured in the **ADR-0023 amendment dated 2026-06-08**
  ("User-verified phase contracts deferred to v2.1"). Read that first.
- It is deliberately **distinct** from two already-rejected
  alternatives and must not be conflated with them:
  - **Alternative B (strict linear wizard)** — rejected for breaking
    freeform navigation. This proposal uses **soft gating** (warn, do
    not hard-lock), preserving freeform.
  - **Alternative E (verification card)** — rejected because
    *automated* thresholds imply surgical-planning-tool-grade
    positioning (ADR-0009 IEC 62366 relaxation). This proposal records
    a **human** attestation, making no automated clinical claim.

## Scope gate — ADR first

This is a **positioning-level** change (research-tool-grade →
potentially surgical-planning-grade), not a UI tweak. It must land an
ADR (new, or an ADR-0023 amendment) **before any implementation**,
resolving the open questions below.

## Open questions for the ADR

1. **Invalidation / staleness.** If upstream data changes after its
   contract was verified, downstream pass state must invalidate (a
   stale ✓ is worse than none). Needs dependency tracking + MRML
   observers on the actual data with a re-lock cascade — not just a
   persisted flag.
2. **Optional-stage DAG.** Encode skip paths (e.g. Stage 3 Auto skips
   Stage 2); a wrong edge locks users out of a valid workflow.
3. **Positioning.** Does human attestation avoid the tool-grade
   repositioning that killed Alternative E, or does any gating signal
[...]

PR #553, opened 2026-07-08
ENH: Compute the distance map on Stage-2 accept + auto-attach to plans

RafaelPalomar:
## Summary

Closes the Stage-2 → Stage-4 hand-off gap from the real-anatomy walkthrough: the distance-map *compute* existed (#542) but nothing triggered it, so Planning opened with no map. Scope note: the full accept/contract machinery (#440) is ADR-gated and stays v2.1; the v2.0 trigger rides the **existing explicit human action** that completes Stage 2 — the canonical import (the same hook the surface representations use).

Two halves, test-first (RED pins → implementation):

1. **`importSegmentationAsCanonical` computes the map** via a new `ensureDistanceMap` seam: reference volume resolved from the Stage-1 `LiverRole='PortalVenous'` tag, resolve-or-create `vtkMRMLVectorVolumeNode` output tagged `DistanceMap`/`Computed` (the v1 selector contract), per-channel signed-Maurer compute at 2× downsampling (full-res 4-channel float of a 512³ CT ≈ 1 GB). Graceful no-op without a reference volume or resolvable channel — the import itself never fails on the map.
2. **`CreateResectionPlan` auto-attaches** the scene's computed map through `SetAndObserveDistanceMapVolumeNode` (ADR-0031: the map lives on the plan wrapper). Untagged vector volumes ignored; empty scene leaves the reference unset as before.

Net: load volume → tag role → import canonical → **Planning opens with anatomy in 3D and the plan's distance map attached**.

## ADR references

- ADR-0023 §Stage 2 hand-off — the canonical import is the stage-completing action; it now produces the downstream inputs.
- ADR-0031 — distance map on the resection-plan wrapper; the auto-attach realises "Planning opens ready".
- ADR-0027 — invariant-test-first (RED commit precedes implementation).

## Conformance

- `test_liversegmentation_distance_map_on_import.py`: import mints exactly one tagged, computed, image-carrying map; re-import reuses the node; no reference volume → import succeeds, no map.
- `test_resection_plan_distance_map_autoattach.py`: computed map reaches the new plan's reference; no map → unset; untagged vector volumes ignored.

## Test plan

- [x] RED first: 3 skip-pending (Python seam) + 1 hard-fail (C++ attach case), degenerate cases already green
- [x] GREEN after: launched run 161 passed / 0 failed (was 157 + 1 failed + 3 pending)
- [x] Both new files skip cleanly under bare pytest (two-harness discipline)
- [x] Incremental build clean (new warnings: none; pre-existing only)

## UX impact

No new UI. The existing "Import as canonical" action now takes a few seconds longer (the compute); the result is that Stage 4 works without manual data plumbing. State-machine unchanged.

Part of #538 (completes it together with the merged compute core). #440 remains v2.1 with its ADR gate intact.

### DIFF
--- a/LiverResections/Logic/vtkSlicerLiverResectionsLogic.cxx
+++ b/LiverResections/Logic/vtkSlicerLiverResectionsLogic.cxx
@@ -62,8 +62,12 @@
 
 // MRML includes
 #include <vtkMRMLScene.h>
+#include <vtkMRMLScalarVolumeNode.h>
 #include <vtkMRMLSelectionNode.h>
 #include <vtkMRMLSegmentationNode.h>
+
+#include <string>
+#include <vector>
 
 // VTK includes
 #include <vtkObjectFactory.h>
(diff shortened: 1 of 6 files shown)


RafaelPalomar (commented):
LGTM

PR #567, opened 2026-07-09
Boundary-guard sweep + slice-projection review follow-ups

### LINKED ISSUE #560: STYLE: Replace per-accessor getattr guards with boundary guards in the LayerDM pipelines
The pipeline modules (LiverBezierSurfacePipeline, ControlPolygonPipeline, ResectogramPipeline, SliceContourPipeline, SliceControlPolygonPipeline, the Representations) guard nearly every node accessor with getattr-with-default. Two legitimate drivers became a blanket habit:

1. the bare-pytest unit layer exercises these modules against stub nodes lacking the wrapped C++ accessors;
2. the interaction/observer entry points are called from C++ and must never raise.

Neither requires per-accessor guards. Proposed sweep (behaviour-preserving):

- **Guard at the boundary, not per accessor**: the entry points C++ calls (observer callbacks, CanProcessInteractionEvent / ProcessInteractionEvent / UpdatePipeline, pipeline-creator callbacks) each get one try/except (or a small never-raise decorator); plain attribute access inside.
- **Richer stubs instead of poorer production code**: unit-layer stubs implement the accessors the code needs; a stub missing one should fail its test, not silently no-op production.
- **Keep _safe_get_* helpers only where absence is a real state** (e.g. display node None mid-teardown); delete the rest.

## Scope inventory (updated after the slice-projection PRs, #564/#565)

Sweep targets — guards on API that is always present in production; only stubs lack it:

- `GetVisibility`-with-lambda-default (BezierPlanningRepresentation, FlattenedSurfaceRepresentation)
- `GetDisplayableNode` (SliceContourPipeline, SliceControlPolygonPipeline)
- `GetControlGridVector` / `GetRows` / `GetCols` on the carrier (ControlPolygonPipeline, SliceContourPipeline, SliceControlPolygonPipeline)
- `GetMTime` (every vtkObject has it)
- `IsA` on the creator's view node (both slice pipelines — becomes the creator-callback boundary guard)
- `RequestRender` / `GetDisplayNode` on `self` (base-class methods on the wrapped `vtkMRMLLayerDMScriptedPipeline`)

**Keep list** — absence is a real, designed state; these survive the sweep:

- the two-namespace resolve of `vtkBezierSurfaceSource` (VTKWidgets module then `slicer`) with the `vtkPlaneSource` fallback in SliceContourPipeline — the wrapped class is genuinely absent in the bare unit layer and its namespace varies by build;
- `SetLocatorRadius` on the injected mapper in FlattenedSurfaceRepresentation — the mapper is a designed injection seam (ADR-0014 §3): the bare layer deliberately injects a generic `vtkPolyDataMapper` without the custom API.

Own-attribute getattr-with-default reads (init gaps, not sweep targets) were already fixed on the source branches: `_slice_display_tags` in ResectogramViewManager and `_edge_cells` in ControlPolygonPipeline are now initialised in the constructors.

Deferred until the in-flight interaction/marker branches merge so the sweep lands in one conflict-free pass.

---
_Authored with assistance from Claude (Anthropic Claude Fable 5)._

### LINKED ISSUE #566: Slice-projection review follow-ups: churn re-attach, ring hide, constants, pins, locator UV routing
Follow-ups from the architectural review of the slice-projection work (the slice contour + slice control polygon PRs), promoted to v2.0.0 per the maintainer's call to not defer in-hand changes. All are small and behaviour-preserving except the last, which is a contract alignment.

## Correctness / robustness

- [ ] **Re-attach the slice-node observer after renderer churn** in `SliceContourPipeline` and `SliceControlPolygonPipeline`: the observer is attached in `SetViewNode`, but `OnRendererRemoved` → `OnRendererAdded` cycles clear node handles and the re-derive path does not re-attach it — a reslice after churn would stop recutting until the next full update. Re-derive in `OnRendererAdded` (the same pattern the display-node handles already follow).
- [ ] **Hide `_ring_actor` when the carrier leaves Planning** in `SliceControlPolygonPipeline`: a stale hover ring can persist one state transition.

## Style

- [ ] **Name the magic numbers**: the 0.78 mid-tone handle base and glyph scales 13/20 (`SliceControlPolygonPipeline`), the 0.6 marker radius factor (`FlattenedSurfaceRepresentation`) — named constants beside the existing `FADE_DISTANCE_MM` block.

## Coverage

- [ ] **Pin `EnsureLocatorNode`'s hidden-display invariant** in `test_ensure_locator_node.py` (the locator display starts invisible; the gesture toggles it).
- [ ] **Pin the MatRatio centre correction at a non-unity aspect ratio** (`FlattenedSurfaceRepresentation` marker placement / `ResectogramPipeline._display_to_uv` — currently only exercised at ratio 1).

## Contract alignment (ADR-0025)

- [ ] **Route the picked (u, v) through the locator node** instead of the direct `ResectogramPipeline` → `FlattenedSurfaceRepresentation.SetPickedUV` thread: ADR-0025 makes `vtkMRMLLocatorNode` the locator-state carrier, and the direct thread hides the marker position from other consumers. Either persist UV on the locator or derive it there; keep the seam test.

Sibling: #560 (getattr boundary-guard sweep) — same files, coordinate so the passes do not conflict.

---
_Authored with assistance from Claude (Anthropic Claude Fable 5)._

RafaelPalomar:
Closes #560, closes #566.

## What

One conflict-free pass over the LayerDM pipelines bundling the getattr boundary-guard sweep (#560) with the slice-projection review follow-ups (#566), plus one demo-found fix.

### #560 — boundary guards instead of per-accessor getattr
- Every C++-called entry point (`UpdatePipeline`, `CanProcessInteractionEvent`, `ProcessInteractionEvent`, `OnRendererAdded`/`Removed`, `OnReferenceToDisplayNodeAdded`, observer callbacks, creator callbacks) now carries one never-raise try/except; plain attribute access inside.
- 79 per-accessor guards removed across the five pipelines + two Representations; unit-layer stubs enriched to implement the accessors the code needs (a stub missing one now fails its test loudly).
- Keep-list retained: the two-namespace `vtkBezierSurfaceSource` resolve, the injected-2D-mapper API probes (ADR-0014 §3 seam), Algorithm-class resolution, and the `_safe_get_*` helpers whose defaults cover a real None-node / unset-field state.

### #566 — review follow-ups
- **BUG**: slice-node observer re-attached after renderer churn in both slice pipelines (reslicing stopped recutting after a remove/add cycle) — pinned red→green in both suites.
- **BUG**: hover ring hides when the carrier leaves Planning — pinned red→green.
- **ENH**: picked (u, v) routed through `vtkMRMLLocatorNode.PickedUV` (transient, (-1,-1) sentinel, same persistence/CopyContent rule as `PickedPositionWorld`); the producer writes it beside the world point and the strip Representation reads it back — the direct `SetPickedUV` seam is gone (ADR-0025: the node is the locator-state carrier). CxxTest + producer + Representation pins.
- **ENH**: coverage pins for `EnsureLocatorNode`'s hidden-display invariant and the MatRatio marker squeeze at a non-unity aspect ratio.
- **STYLE**: `HANDLE_MIDTONE_FACTOR`, `HANDLE_GLYPH_SCALE_PX`/`RING_GLYPH_SCALE_PX`, `_MARKER_RADIUS_FACTOR` named beside `FADE_DISTANCE_MM`.

### Demo-found fix
- **BUG**: re-entering the module left the resectogram strip black until a manual redraw — `showEvent` now kicks one deferred render (pinned red→green).

## Validation

- Bare harness: 95 passed, 0 failed.
- Launched harness: 291 passed, 15 known skips, 1 known xfail, teardown-crash-free.
- CxxTests: 24/24 passed (including the new `PickedUV` pins).
- Interactive demo pass on real anatomy: strip marker gesture, control-point drag in 3D + slices (13.6 ms/event), reslice after drawer re-open, hover highlights, and module re-entry all verified.

---

### DIFF
--- a/LiverResections/LiverResectionsLib/ControlPolygonPipeline.py
+++ b/LiverResections/LiverResectionsLib/ControlPolygonPipeline.py
@@ -197,9 +197,7 @@ class ControlPolygonPipeline(_PipelineBase):
         self._display_node = displayNode
         self._data_node = None
         if displayNode is not None:
-            getter = getattr(displayNode, "GetDisplayableNode", None)
-            if getter is not None:
-                self._data_node = getter()
+            self._data_node = displayNode.GetDisplayableNode()
             self._attach_observer(displayNode)
             if self._data_node is not None:
                 self._attach_observer(self._data_node)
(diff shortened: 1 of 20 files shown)


RafaelPalomar (commented):
LGTM

Issue #578, opened 2026-07-09
Stage 4: camera+geometry-aware resection initialization (port the v1 contour-init flow)


RafaelPalomar:
**E2E stabilization walkthrough finding (2026-07-09, pure-AI pass).**

## Symptom

"Place resection" produces a default 4×4 grid centred at the origin — no relation to the liver, no interaction. In v1, initialization was camera- and geometry-aware: the surgeon placed a contour ON the liver surface (projected from the camera view) which then seeded the Bezier resection surface. The v2 default is not a usable initialization, and it admits no interaction at all; per the maintainer it must follow the same interaction/visualization pattern as the resection control points.

## v1 mechanics (for parity)

- The resections widget built a hidden **target organ model** from the parenchyma segment's closed-surface representation and set it on the resection node (`SetTargetOrganModelNode`, opacity 0, `HideFromEditors`).
- The init markups (slicing contour / distance contour) interacted with that target mesh, camera-aware; the resulting ring on the liver surface seeded the Bezier fit that became the Planning surface.

## v2 status (the thread is half-built)

- `InitMode` lives on the carrier (ADR-0019); `SlicingPlaneInit` / `DistanceSpheroidInit` Representations exist and are state-dispatched (ADR-0013 §4).
- The interactive placement routing exists but its launched test skips ("no realized 3D view — ProcessInteractionEvent needs a renderer/camera"; verified on `:0` only for the placement gestures, not the full flow).
- The **commit-boundary ring extraction is a strict xfail**: `TODO(T2-target-mesh-weakref)` — the Init Representation's consume sites never feed the extractor the weakref'd target mesh (ADR-0014 §1). The Algorithm library already carries the ring extractors and `vtkLiverBezierFitter`.
- `CreateResectionPlan` seeds the default origin grid — the placeholder the walkthrough saw.

## Scope

1. **Target-organ hand-off**: resolve the target mesh from the canonical liver segment (Stage-2 → Stage-4 hand-off; the ADR-0014 §1 weakref'd `GetTargetModelNode` contract) when a plan is created/selected.
2. **Camera-aware init interaction**: the init contour/points placed against the target mesh from the current camera, interactive with the same visualization/interaction pattern as the resection control points (the ADR-0032/0033 pipeline seam — hover/grab, glyph language).
[...]

PR #580, opened 2026-07-10
Resection init slice 1: wire the liver target model to the carrier

RafaelPalomar:
First slice of #578 (camera+geometry-aware resection initialization).

## What

- **Lifts the `T2-target-mesh-weakref` strict xfail** — the pin's fixture never gave the carrier a display node, so the pipeline could not bind the carrier and the test measured the fixture gap; `commit()` has consumed the weakref'd target through `GetTargetModelNode()` since the seam landed. Fixture fixed, pin passes on the existing wire.
- **Builds the genuinely missing wire**: `ensure_target_model` resolves the Stage-2 canonical segmentation's SCT-tagged liver segment, mints the hidden target-organ model from its closed surface (the v1 pattern — deep-copied polydata, hidden from editors, never rendered, resolve-or-mint singleton) and attaches it to the plan carrier's weak target reference (ADR-0014 §1). Place resection runs it after `CreateResectionPlan`; no canonical liver → graceful no-op. The cross-module contract is scene data (role attribute + SCT tag), not a Python import.

With this, the commit-boundary ring extraction finally has a liver to cut. Next slices per the v1 study: plane derivation from the placed points, placement routing + draggable handles on the control-point pattern, PCA grid seeding on commit (kills the origin-grid symptom), live shader contour.

## Validation

Six pins green (three new target-model invariants + the two commit-boundary pins incl. the former xfail); full launched sweep 296 passed / known skips only.

---

### DIFF
--- a/LiverResections/LiverResectionsLib/CMakeLists.txt
+++ b/LiverResections/LiverResectionsLib/CMakeLists.txt
@@ -32,6 +32,7 @@ set(LiverResectionsLib_PYTHON_SCRIPTS
   ResectogramPipeline.py
   SliceContourPipeline.py
   SliceControlPolygonPipeline.py
+  TargetModel.py
   ResectogramViewManager.py
   Representations/__init__.py
   Representations/BezierPlanningRepresentation.py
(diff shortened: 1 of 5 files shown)


RafaelPalomar (commented):
LGTM

PR #582, opened 2026-07-10
Resection init slice 3: auto-seeded, drag-editable slicing-plane handles

RafaelPalomar:
Third slice of #578 — the first *visible* one.

## What

- **Auto-seed** (v1 parity, camera-aware): entering Init+SlicingPlane with a target mesh attached seeds the two handles across the target's bounds along the **camera's right vector** — the initial bisector plane cuts vertically through the surgeon's view and the handles sit left/right on screen, maximally draggable. Headless fallback: the world x axis (v1's exact default). One-shot, with a re-entrancy latch (seeding mutates the carrier, whose ModifiedEvent re-enters the reconcile hook before the counter increments — caught live as a counter unwinding to 194; the LayerDM observer-mutation recursion again).
- **Drag-editable handles on the control-point grammar** (ADR-0033 pattern): press within the pick radius claims with the real squared display distance for LayerDM arbitration; moves relocate the grabbed handle on the camera focal plane and **re-derive the bisector plane live** (slice 2's kernel); release drops the grab. Bare hovers stay unclaimed — camera interaction untouched. Outside Init the pipeline keeps declining (the Planning drag remains ControlPolygonPipeline's, ADR-0033).
- The legacy fill-placement path survives as the no-target fallback.

## Validation

Four new pins (seed positions + derived plane + idempotence; no-target no-op; full press/move/release gesture with live re-derivation; Planning-declines guard); full launched sweep **303 passed** / known skips; bare harness 95 passed.

**This is the live-test checkpoint** for slices 2+3 together: place a resection → handles appear straddling the liver facing the camera → drag them → the plane follows. Builds on #580/#581 (merged).

---


RafaelPalomar:
## /slicer-review — spec-graded review (ADRs + architecture docs as the spec)

**Spec status:** graded against 35 ADRs + 10 architecture docs, including this PR's own ADR-0035 (checked for internal consistency with ADR-0019/0032/0033).

**Conformance:** all 20 files graded *advance*, none *drift*. Single-writer discipline verified by grep: the only production `SetState` for the Init→Planning transition lives in `ResectionStateMachine.py`; zero phase-attribute writes outside the machine. No custom displayable manager (ADR-0013 upheld); interaction stays on the ADR-0032 pipeline seam.

**Findings — all fixed on this branch (commit "Harden the init loop per the ADR-graded review"), each pinned red→green:**

1. *(blocker)* Two composite pins imported `LiverResectionsLib` unguarded at test scope — the file's bare-pytest CTest row errored instead of skipping (the two-harness trap at row granularity; confirmed empirically before the fix).
2. *(should-fix)* Both pipelines' digest-gated render keys omitted the ADR-0035 phase — a phase flip raised outside a local gesture reconciled the composite without a repaint. Now keyed on `phase_token()`.
3. *(should-fix)* A raising re-fit inside `request(PlaneHandleDropped)` stranded the phase in-flight; a raise now reads as a failed fit and the drop always lands a resting phase.
4. *(should-fix)* Scene-reload clobber: adopting a carrier with persisted init handles re-seeded them over the surgeon's adjustments. The auto-seed now adopts per-slot node-side evidence — and fixing it surfaced a reentrancy bug (the placement kernel must claim its slot before `SetSlicingPlaneInitPoint`, whose synchronous `Modified` re-enters the adoption path).
5. *(coverage)* Restored the dropped corollary pin: the `SurfaceGrabbed` commit performs zero ring extractions (pure state flip, ADR-0035 §4).
6. *(style)* Stale `commit()`-era docstrings updated to the drop boundary; the wrapping-pin test gained its per-file bare CTest row.

**Deferred to follow-ups (post-merge acceptable):**
- Architecture-diagram refresh: `resection-state-machine.md` (flat Init) and `rendering-pipeline.md` ("picks one active at a time") now trail the ADR-0035 interior + composite dispatch.
- Transform-aware contour uniform: the shader band assumes the target model sits under an identity transform (world RAS pushed into a model-coordinate uniform).
- ADR-0035 [review] note on the pre-existing C++ mirror arm in `vtkSlicerLiverResectionsLogic` (test-pinned, not a runtime driver) + stale `commit()` comments in that file (left untouched to avoid whole-file clang-format churn on this branch).
- DistanceSpheroid extraction boundary has no caller/pin post-`commit()` — ADR-0035 marks it [future]; worth an explicit issue when that mode is taken up.
- Gesture-grammar dedup is already tracked in #583.

**Verification after fixes:** launched harness 336 passed / 15 known skips, zero crashes; bare harness 111 passed / 38 skipped; per-file bare CTest rows for all changed launched test files skip cleanly.

**Verdict: Ready** — no open blockers; live behaviour maintainer-verified (composite loop + state-machine regression pass).

---
*Review and fixes drafted by an AI assistant (Claude Fable 5); posted on behalf of the maintainer.*

RafaelPalomar (commented):
LGTM

Issue #574, opened 2026-07-09
Stage 2: anatomy segments table + tool registry + background job queue (ADR-0034)


RafaelPalomar:
Implements ADR-0034 (Stage-2 anatomy segments table, tool registry, background job queue) — the redesign committed after the 2026-07-09 end-to-end walkthrough. Read the ADR first; this issue is the increment plan.

## Increments (each PR-sized, test-first per ADR-0027)

1. **Table, read-only.** The pre-seeded anatomy segments table (structure rows, status vocabulary `○ Missing / ⟳ Running / ● Review / ✓ Confirmed / ∅ Marked absent / ✎ Interactive…` as glyph + text) rendered from the current canonical-node state, alongside the existing tabs. Pins: pre-seeded empty state; row status derivation.
2. **Row actions + retire the tabs.** Run/Confirm/Discard/Edit/Mark-absent move into rows; per-segment confirm tag lands (scene-persistent, beside the SCT tag); results land in canonical as `● Review`; re-run/edit demotes to Review. The QTabWidget, the cards, and the separate import section are removed. Completion predicate flips to all-rows-resolved; shell predicate + downstream confirm-tag filtering + test migration ride along.
3. **Import unification.** Importing a loaded segmentation populates rows via LabelToSCT name-matching; unmatched segments get an in-row structure combo; imports arrive as `● Review`.
4. **Tool registry.** Structure SCT → ordered chains (default first; wrapper + task + labels + modality flag + preferred input role). `INFERENCE_TARGETS` promotes into it; per-row ▶ becomes a split-button over the row's chains.
5. **Job queue.** `SegmentationJobQueue`: main-thread `qt.QProcess`, sequential, coalescing on `(task, input volume)`, signals feeding row status, cancel (✕) kills the running job, teardown kills children. Replaces the blocking readline + processEvents runner (explicitly named as the superseded pattern in ADR-0034 §5 — keep the command builder). Observe the PythonQt GC/signal discipline from the ADR; conformance-review item.
6. **"Segment anatomy" macro button** enqueueing all Missing rows (two coalesced backend calls for the four structures).
7. **Validate-and-next seam.** Module exposes the completion predicate + explain API; failed validation marks unresolved rows (red background + glyph + summary line, clearing live). Shell-side button wiring lands with the phase-contracts work; the seam ships here.

## Relationships

[...]

PR #589, opened 2026-07-14
ENH: Import via explicit correspondence table through the landing kernel (#574 increment)

RafaelPalomar:
Implements the **import-unification increment of #574** (ADR-0034 §Decision 2, "the import path unifies"), redesigned mid-review per maintainer live-testing: import goes through an **explicit correspondence table**, not auto-matching.

## What

The legacy *Load segmentation* section (a second grammar whose promote-the-source semantics conflicted with the pre-seeded canonical) is retired. In its place, one toolbar **Import…** gesture:

- The picker dialog carries a mapping table — one row per **source** segment, an *Import as* combo per row offering the four expected structures. **Every row defaults to "— skip —"**: no auto-match of any kind (no name bridges, no tag routing); the surgeon states each correspondence explicitly. Many→1 is natural (several rows → Tumors → N Mass segments).
- **Mapped segments land as the expected structure** — vocabulary title (numbered for multifocal / extra rows), visual defaults, SCT tag — through the same landing kernel as the AI path, as `In progress` + source `imported`, reviewed with the same status-cell confirm.
- Structures already landed show an "(already present)" suffix; picking one lands an **extra row** — a landed/Confirmed row is never overwritten (maintainer-agreed rule).
- The source node is **consumed only when every segment was mapped**; partial imports keep it (status label says so); all-skip is a graceful no-op.
- Canonical identity is stable throughout; distance-map + surface triggers ride the path.

## Hardening found by live-testing during review

- The flags-taking `QDialogButtonBox` constructor doesn't marshal through PythonQt → buttonless dialog; built via the `standardButtons` property, pinned.
- `setParent(None)` + `deleteLater()` double-freed the dialog on the next event-loop spin (PythonQt takes ownership of parentless widgets); it stays parented, pinned by a staged repro.
- Lambda signal connections are GC'd by PythonQt (bound-method slots); `qMRMLNodeComboBox` emits no `currentNodeChanged` between identically named nodes — the table records its populated node id and a mismatch reads as all-skip.

## Verification

Full launched sweep, canonical four-root order: **511 passed / 16 skipped / 0 failed**, no crashes; bare rows skip cleanly; ruff clean. Dialog shape (Ok/Cancel wired, skip defaults, repopulation, identity landing, consumption rules) fully pinned headless; the modal `exec_()` itself eyeballed live.

Remaining #574 increments: tool registry + Run ▾ split-button, validate-and-next explain API.

---
*Drafted by an AI assistant (Claude Fable 5); reviewed and submitted on behalf of the maintainer.*


RafaelPalomar (commented):
LGTM

Issue #569, opened 2026-07-09
LiverSegments LayerDM display-node migration


RafaelPalomar:
Migrate LiverSegments' display side onto the LayerDM Pipeline pattern (ADR-0013): C++ data-only display node + Python Pipeline + Representations, replacing the legacy display path. Data-model/algorithmic work already landed in v2.0.

Part of the v2.0.0 scope re-baseline (2026-07-09): previously deferred to v2.1 by ADR-0012, absorbed into v2.0.0 per the ADR-0012 Amendments (PR #568). Tracked on #305.

---
_Authored with assistance from Claude (Anthropic Claude Fable 5)._

RafaelPalomar:
## Closing: Stage 3 (the VMTK feed) landed — verified against the tree

Triage sweep against `preview` @ `4421370`. The ADR-0037 staging is complete; this issue has been overtaken by its own PRs.

**Evidence**
- Stage 1 — annotation carrier + placement pipeline: #592
- Stage 2 — annotation table + arming seam: #593
- Stage 3 — the VMTK feed: #597, then #601 (territory-map compute on the carrier), #602 (per-territory Place + composite tree panel), #603 (multi-system seeding + centerline extraction)
- Post-merge fixes: #596, #599, #600, #607

**Verified in-tree**
- `VascularTerritories/VascularTerritoriesLib/TransientVmtkSeeds.py` — the pure seed-payload core; the `ExtractCenterline` input is built off `vtkMRMLCustomTerritoriesNode`, not Slicer markups (ADR-0037 §Decision 4).
- `VascularTerritories/Testing/Python/test_territories_vmtk_feed.py` — 11 tests pinning the feed, including the >=2-seed gate and re-extraction idempotency.
- Graceful degradation is real: the action disables with an explanatory tooltip when `ExtractCenterline` is absent (`VascularTerritories.py:991-1015`).
- The module rename `LiverSegments/` -> `VascularTerritories/` (ADR-0023) is complete; only an untracked stale `__pycache__` entry remains on disk.

No code gap remains. Reopen if the eyeball pass finds behaviour this misses.

---
_Triaged with assistance from Claude (Anthropic Claude Opus 5)._

PR #600, opened 2026-07-16
ENH: Modernise the VascularTerritories panel to Python-widget composition (#569 slice 2)

RafaelPalomar:
Slice 2 of the VascularTerritories widget-modernization — completing ADR-0037 §Decision 3. Stage 2 (#593) composed the new `TerritoriesTableWidget` into the panel but left the entire v1 operating surface wired alongside it, so the panel presented two incoherent flows (surfaced in an eyeball: seeds place onto the carrier but the visible v1 buttons/selectors drive the old scene-attribute model).

## What changed
Retire the v1 widgets that merely duplicate the table (and the per-segment input selector), re-home the rest into a Python-composed panel (ADR-0004, the `ResectionPlanningWidget` precedent):

- **Retired** (widgets + handlers/helpers): `ColorPickerButton`, `showHideButton`, `addSegmentationButton`, `SegmentsWidget` (per-territory colour/label/visibility now live in the table), `inputSegmentSelectorWidget` (per-segment input — extraction runs on the whole input segmentation), `vascularTerritoryId`; `onColorChanged`, `onShowHideButton`/`refresh…`, `onAddSegmentationButton`, `updateVascTerrList`, `createCenterlineNode`, `copyIndex`, `mergePolydata`, `getDisplayNodeAndSegmentId`, the dead `_registerVesselHighlightPipeline`, and the retired-selector parameter-node sync.
- **Kept + re-homed**: `inputSurfaceSelector` (feeds `pickSurface` + extraction surface), `SegmentationShow3DButton`, the composed table, the **Extract centerlines** action.
- **Kept functional for slice 3** (untouched this slice): the **Compute territory map** action + `selectedVascularTerritorySegmId` + `calculateVascularTerritoryMap`.

## Scope
This is the panel retirement only. Slice 3 re-sources `build_centerline_model` off the scene-scan onto the carrier's `CenterlineRefs`, derives the map-output target (retiring `selectedVascularTerritorySegmId`), and amends ADR-0037 — that's the slice that makes the seed→extract→map flow fully coherent end-to-end.

## Tests
New `test_territories_widget_panel.py` (11 launched invariants, exact retire/keep name contract): retired widgets/handlers absent, kept surface present + wired, panel builds cleanly, map-path guard, teardown clean. The Stage-1 highlight-wiring invariant stays green.
- Launched (VascularTerritories root): 71 passed / 1 skipped (only the env-gated real-VMTK test), no crash.
- Regression (VascularTerritories before LiverSegmentation): 194 passed / 2 skipped, no crash.

---
*Authored with the assistance of Claude (Anthropic Opus 4.8). Reviewed by the maintainer before merge.*

### DIFF
--- a/VascularTerritories/Resources/UI/VascularTerritories.ui
+++ b/VascularTerritories/Resources/UI/VascularTerritories.ui
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ui version="4.0">
- <class>SegmentsWidget</class>
- <widget class="qMRMLWidget" name="SegmentsWidget">
+ <class>VascularTerritoriesWidgetRoot</class>
+ <widget class="qMRMLWidget" name="VascularTerritoriesWidgetRoot">
   <property name="geometry">
    <rect>
     <x>0</x>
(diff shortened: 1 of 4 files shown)


RafaelPalomar (commented):
LGTM

PR #601, opened 2026-07-17
ENH: Wire the territory-map compute onto the annotation carrier (#569 slice 3)

RafaelPalomar:
Slice 3 (final) of the VascularTerritories widget-modernization — completes ADR-0037 §Decision 3/4. The panel and placement were modernized in slices 1–2, but the territory-map compute was still wired to the retired v1 scene-attribute model, so placed seeds didn't flow to the map. This re-sources the whole compute path onto the annotation carrier and makes the seed → extract → map flow coherent end-to-end.

## What changed
- **New pure `TerritoryLabelMap`** (`territory_label_ints`): deterministic string→int labelmap scalar per territory (annotation order → `index+1`; 0 = background). The int is an arbitrary distinct label the colormap colours — explicitly NOT an SCT code (ADR-0011). Registered in `_PYTHON_SCRIPTS`.
- **`build_centerline_model` re-sourced onto the carrier**: iterates the carrier's `CenterlineRefs` + `Groupings` (not `getNodes("*Territory*")`), marking each centerline with its derived int. The old `SegmentationId` filter collapses (one carrier = one map).
- **Derived map target** (`ensureTerritoryMapOutput`): the output `vtkMRMLSegmentationNode` is created/reused via a new carrier node-reference role `TerritoryMapOutput`, with a per-carrier ordinal `SegmentationId` (scene `max+1`, collision-free). The C++ `calculateVascularTerritoryMap` signature is unchanged.
- **Retired `selectedVascularTerritorySegmId`** (selector + param-node role + `.ui`); `onCalculateVascularTerritoryMapButton` now resolves input from `inputSurfaceSelector`, centerline + target from the carrier. Compute-button enablement re-sourced onto the `InputSurface` role.
- **ADR-0037 amended**: the retired-widget set (slices 2+3), the two-step flow (place seeds → Extract centerlines → Compute territory map), and the derive-target-from-carrier decision.

## Tests
`test_territories_map_compute.py` (12 invariants: label-map determinism [3 bare] + carrier-sourced `build_centerline_model` + derived/reused target + distinct per-carrier ordinals + selector-retired + button resolves without the selector). The slice-2 guard flipped to assert the selector's removal. Real C++ map run stays eyeball-gated (stubbed in tests).
- Launched (VascularTerritories root): 82 passed / 1 skipped (env-gated VMTK). Combined: 205 passed / 2 skipped, no crash. Highlight-wiring + slice-2 panel guards green.

Completes #569 (the VascularTerritories off-markups transition, end to end).

---
*Authored with the assistance of Claude (Anthropic Opus 4.8). Reviewed by the maintainer before merge.*

### DIFF
--- a/Docs/adr/0037-vascular-territories-off-markups.md
+++ b/Docs/adr/0037-vascular-territories-off-markups.md
@@ -166,3 +166,60 @@ modernisation; (3) the VMTK feed + graceful degradation.
   consumer of the vessel highlight.
 - [future] The Auto/Couinaud path modernisation, if ever taken off its
   current stock display.
+
+## Amendment — the territory-map compute path off the carrier (slices 2 + 3)
+
+The staged transition (§Staging) completed the panel-modernisation half in
+two increments beyond §Decision 3.  This amendment records the compute-path
+decisions the implementation crystallised, so the retired-widget set and the
+final workflow are documented in one place.
+
+### Retired-widget set across slices 2 + 3
+
+Slice 2 retired the table-duplicating v1 surface (`ColorPickerButton`,
+`showHideButton`, `addSegmentationButton`, `SegmentsWidget`,
+`inputSegmentSelectorWidget`, `vascularTerritoryId`) and their handlers, plus
+the dead `_registerVesselHighlightPipeline` and the `copyIndex` Logic helper.
+**Slice 3 additionally retires `selectedVascularTerritorySegmId`** — the
+per-map output-segmentation selector — together with its `.ui` widget, its
+`VascularTerritorySegmentation` param-node role, and its `setup()` wiring.
+The Compute-territory-map button and `onCalculateVascularTerritoryMapButton`
+survive; only the output *selection* is gone.
+
+### Coherent two-step compute flow
+
+The surviving workflow is a coherent two-step sequence over one input
+surface: **place seeds** (add-on-click into the carrier, per §Decision 2) →
+**Extract centerlines** (`onAddCenterlineButton` → `extractCenterlines`, the
+transient-markups VMTK feed of §Decision 4) → **Compute territory map**
+(`onCalculateVascularTerritoryMapButton`).  The compute action gates on an
+input surface being selected rather than on a separate output selector.
+
+### Derive the map target from the carrier
+
+The output map target is **derived from the carrier, not selected** — a
+carrier node-reference role (`TerritoryMapOutput`) resolves to exactly one
+`vtkMRMLSegmentationNode`, auto-created and attached on first compute and
+reused thereafter (one carrier == one map, consistent with the
+`CenterlineRefs` + `Groupings` map of §Decision 4).  Dropping the selector
+removes a manual step and the class of "wrong output picked" error; it also
+retires the redundant `int(...)` string-tag reader the old
+`slicer.util.getNodes("*Territory*")` scene scan depended on, because
+`build_centerline_model` now sources the carrier's `CenterlineRefs` directly.
+
+### Arbitrary-int labelmap scalar (explicitly not an SCT code)
+
+`vtkSlicerVascularTerritoriesLogic::MarkSegmentWithID` stamps a per-territory
+integer into each centerline's `segmentId` point-scalar, and
+`calculateVascularTerritoryMap` reads + re-stamps a per-map
+`VascularTerritories.SegmentationId` ordinal on the output.  Both are
+**arbitrary distinct positive labelmap scalars, NOT SCT terminology codes**
+([ADR-0011](https://github.com/ALive-research/Slicer-Liver/blob/preview/Docs/adr/0011-sct-terminology-dispatch.md)
+reserves SCT for the accepted-plan terminology surface, not this internal
+watershed label).  The transition
+derives them deterministically: the per-territory int is
+`index + 1` over the carrier's `GetAnnotationTerritoryIds()` order (0 is the
+labelmap background), so the same territories derive the same ints across
+repeated calls (the pure `VascularTerritoriesLib.TerritoryLabelMap` core); the
+per-map ordinal is issued one past the maximum `SegmentationId` already
+stamped in the scene, so two carriers never collide.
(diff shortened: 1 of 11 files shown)


RafaelPalomar:
## /slicer-review report (pre-Ready gate)

_Run by liver-reviewer agent on 2026-07-25._

**Spec status:** found 10 architecture docs + full ADR set; graded against
ADR-0037 (§Decision 4 + the compute-path amendment = spec of record), 0011,
0014, 0004, 0027, 0023/0024. Scope: 8 commits, +1635/-110, base `preview`,
head `feature/vasc-territories-slice3-compute`. CI green on head.

### Blockers

None. The four fixes each land with a red→green invariant test that pins the
*specific* failure it repairs (ADR-0027), and the compute-path rewire
conforms to ADR-0037 §Decision 4.

### Architecture conformance (advances the target)

- **`build_centerline_model(carrier, colormap)`** — re-sourced off the
  carrier's `CenterlineRefs` + `Groupings` map (`getCenterlineReferenceIDs`
  / `GetGrouping`), retiring the `slicer.util.getNodes("*Territory*")` scene
  scan and the collapsed `SegmentationId` int filter. Matches ADR-0037
  §Decision 4 "one carrier == one territory map == all its centerlines".
- **Retired output selector** — `selectedVascularTerritorySegmId` and its
  `VascularTerritorySegmentation` param-node role dropped from the selector
  list, `.ui`, and `setup()` wiring; output derived via the new
  `TerritoryMapOutput` carrier node-reference role (ADR-0014 wrapper/carrier,
  no new node family). The ADR-0037 amendment (42d2369) documents the
  retirement + the two-step flow before the code relies on it — correct
  ADR-first discipline.
- **`TerritoryLabelMap.py`** — pure `territory_label_ints` core lives in Lib
  (ADR-0004 Python side), bare-unit-testable, deterministic `index+1` over
  `GetAnnotationTerritoryIds()`. The docstring correctly states the int is an
  arbitrary watershed label, NOT an SCT code (ADR-0011 boundary respected).

### Correctness (MRML / VTK)

- **triangulate-before-decimate (15de967)** — `preprocessAndDecimate` now
  runs `vtkTriangleFilter` (PassLines/PassVertsOff) ahead of `vtkDecimatePro`,
  fixing the strip→empty-mesh bug. Full `vtkSmartPointer` discipline through
  triangulate→decimate→clean→normals→`ShallowCopy`; no leaks, no raw new.
- **real-segment resolution (80432b3)** — `polyDataFromSegmentation` converges
  on `VesselHighlightWiring.closed_surface_polydata` (whole-vessel-tree append)
  instead of feeding a territory-semantic id into
  `GetClosedSurfaceRepresentation` (which resolved no segment → 0-point mesh).
  Also fixes the bare `logging.error` statement (was a no-op reference) to a
  real call + `return None`.
- **`<2`-seed skip (4090aa5) + None/empty-surface guard** — under-seeded
  territories are skipped (a centerline needs ≥2 endpoints), and
  `_extractOneTerritory` early-returns on a None/empty surface rather than
  feeding the real VMTK extractor a mesh it hard-fails on. Honest degradation,
  matches the amendment.
- **liver-guard (8032c71)** — `GetLiverSegmentId` matches by SCT token
  `SCT^10200004` in the `TerminologyEntry` tag (ADR-0011), not by name;
  `calculateVascularTerritoryMap` now returns early with a legible
  `vtkErrorMacro` when no liver segment resolves, instead of exporting an
  empty labelmap and a silently blank map.
- **transient-fiducial lifecycle** — `_extractOneTerritory` tears the seeds
  node down in a `try/finally RemoveNode`; no persisted markups (ADR-0014/0037).
- **`Reset()`/attribute round-trip** — `SegmentationId` is copied to
  `segmentationIdCopy` before `Reset(nullptr)` clears the attribute table and
  re-stamped after; correct.

### Correctness — non-blocking (pre-existing, this PR touches the function)

- `calculateVascularTerritoryMap` (logic .cxx ~L362): the `!mrmlScene` guard
  logs `vtkErrorMacro` but does **not** `return`, so the next line
  dereferences the null `mrmlScene->AddNewNodeByClass`. Same shape at
  `preprocessAndDecimate` (~L436). Pre-existing v1 hazard, unreachable on the
  real path (scene always attached), but the PR edits both functions — a
  one-line `return` after each macro would close it. Recommend a follow-up,
  not a gate.
- `labelmapVolumeNode->GetDisplayNode()->SetAndObserveColorNodeID(...)`
  (~L407) dereferences `GetDisplayNode()` without a null check. Same
  non-blocking category.
- `labelmapVolumeNode->Modified();` retains a "Is this enough?" comment — the
  labelmap is imported to the segmentation immediately after, so this is
  cosmetic. Consider resolving the comment.

### Style

- No naming / macro / class-separation deviations. C++ stays on the logic
  (decimation/classification per ADR-0004), Python owns the VMTK feed + widget
  composition. Copyright headers present on the new files.

### Coverage (ADR-0027 / ADR-0008)

Genuinely pins the fixes:
- `test_territories_surface_resolution.py` — `test_preprocess_tolerates_triangle_strip_surface`
  builds a real strip fixture (`vtkStripper`, asserts `GetNumberOfPolys()==0`)
  and asserts the decimated result is non-empty — pins triangulate-before-decimate
  directly. `test_polydata_from_segmentation_resolves_a_non_empty_mesh` +
  `test_empty_surface_is_not_fed_to_the_extractor` pin the real-segment
  resolution and the None/empty skip (capturing-extractor asserts surface has
  >0 points).
- `test_territories_map_compute.py` — label-map determinism/1-based/distinct
  (bare), carrier-sourced `build_centerline_model` marking each with the
  derived int, `ensureTerritoryMapOutput` first-derive + second-reuse +
  two-carrier-distinct-ordinal, selector retirement, and button-resolves-
  without-selector. The C++ map run itself is stubbed (eyeball-gated) —
  consistent with ADR-0027's env-gated real-VMTK boundary; the eyeball fixes
  are what these tests lock in.
- Both files registered as bare `pytest_scaffold` rows in CMake with the
  import/hasattr SKIP-PENDING guards; launched self-test opts into
  LayerDM + SlicerVMTK module paths. Both harnesses covered.

The `<2`-seed threshold is pinned indirectly (via the extractor-feed tests
and the vmtk-feed suite) rather than by a dedicated count-boundary case;
minor gap, not a blocker.

### Recommendation

READY — no architectural-conformance flags, MRML/carrier invariants hold, the
four eyeball fixes each ship a specific red→green invariant test (ADR-0027),
ADR-0037 was amended before the code leaned on it, and both test harnesses are
wired. CI green. Suggested (non-blocking) follow-ups: add the missing `return`
after the two null-scene `vtkErrorMacro`s and the `GetDisplayNode()` null
check while the compute function is warm; resolve the stale `Modified()`
comment. None gate Ready.

_This PR body should confirm its Conformance section points at these invariant
tests before the toggle is flipped — verify present._

RafaelPalomar (commented):
LGTM

PR #607, opened 2026-08-14
ENH: VascularTerritories usability batch (eyeball round)

RafaelPalomar:
## Summary

Usability batch for VascularTerritories, driven by a live eyeball session on
real anatomy after the ADR-0038 migration. Every fix was verified live on a
real GL display in addition to the test suites.

- **Hover cue restored on placement** — the territory placement pipeline's
  bare-move decline now requests a render, so the adhering marker/halo shows
  while placing (the ADR-0038 extraction had dropped the flush; a regression
  test pins it).
- **Territory-level delete** — the carrier gains `RemoveTerritory` and each
  territory row a Remove control, so an empty territory is no longer
  undeletable; deleting the active territory disarms placement.
- **Affirmative requirements messaging** — Extract centerlines / Compute map
  gained a live "what's missing" line + tooltips (input, SlicerVMTK, seeds,
  reference volume, centerlines) instead of silently disabled buttons.
- **Seed-drag performance** — the seeds table defers its full rebuild while a
  drag is in flight (a `Grabbing` flag on the shared display node, set by the
  pipeline), removing the per-frame tree rebuild that made editing crawl.
- **Centerline visibility follows the territory** — hiding a territory now
  hides its extracted centerline(s) too (gated on territory AND structure
  visibility, associated via the carrier's groupings).
- **Per-territory review status + derived edit-lock** (ADR-0037 amendment) —
  a click-to-cycle status cell reusing Slicer's segment-status vocabulary
  (NotStarted/InProgress/Completed/Flagged); `Completed` derives a lock that
  refuses placement/drag/delete with explaining tooltips, and a geometry edit
  demotes to InProgress. Status round-trips through storage.

Docs: the ADR-0037 amendment + the protection/validation design study, both
in the toctree.

Tests: carrier, table, pipeline, and enablement invariants extended in both
harnesses; the launched suites are the behaviour gate in CI.

---

*Drafted with the assistance of Claude Code (model: claude-opus-4-8);
reviewed by the author.*

### DIFF
--- a/Docs/adr/0037-vascular-territories-off-markups.md
+++ b/Docs/adr/0037-vascular-territories-off-markups.md
@@ -386,3 +386,94 @@ the 3D vessel render, not the territory palette), paired with the structure name
 - [review] Re-extraction is idempotent: a territory's prior centerlines are
   cleared once before its structure loop, so re-running replaces rather than
   accrues.
+
+## Amendment — per-territory status + derived edit-lock
+
+- **Status:** Accepted
+- **Date:** 2026-07-30
+- **Cross-references:** [ADR-0034](0034-stage2-segments-table.md) (the shared
+  status vocabulary + demote-on-edit rule), [ADR-0010](https://github.com/ALive-research/Slicer-Liver/blob/preview/Docs/adr/0010-accessibility-and-i18n.md)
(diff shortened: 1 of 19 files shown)


RafaelPalomar (commented):
LGTM

PR #608, opened 2026-08-14
ENH: LiverVolumetry seed workbench rework (carve model + v2 interaction)

RafaelPalomar:
> **Stacked on #607** (territory usability batch). Do not merge before it;
> after #607 merges this rebases onto preview with the usual drill.

## Summary

Ground-up rework of the LiverVolumetry workflow, panel, and seed interaction,
built through iterative live eyeball sessions on a synthetic overlapping-
segment phantom with analytically known volumes, and hardened by a dual
design review (UX heuristics + Slicer-conventions, verified against Slicer
5.8 source).

### Workflow + panel
- Data-first panel: one input selector, a single primary **Compute volumes**
  action, requirements/status line; reference-volume, duplicate segment
  picker, and output-table selectors removed (module-owned "Volumetry"
  results table, replace-on-compute). Surgeon-term columns
  (Region | Volume (mL) | % of total) and an explicit **Total** row naming
  the denominator. Neutral vocabulary (leftover segment = "Unassigned";
  section = "Total volume"). Seeds are a first-class way to pick regions;
  resections are an optional refinement (no silent no-op branches — gating
  tells the truth).

### Seed model
- **Grouped volumes** mirroring the territories model: named volumes with
  per-volume colour/label, placement arms per volume, seeds nest under their
  volume, one results row per volume.
- **Visibility-composed carve**: visible segments carve each other
  (top-visible owns the voxel); each seed records its visibility snapshot as
  its reproducible definition; compute aggregates carved regions (verified
  against the phantom's analytic volumes).
- **Seed→structure capture**: a placed seed binds to the top visible touched
  segment (layer-ordered), retargetable; per-seed (segmentation, segment)
  binding persists.

### Interaction (post-review v2)
- Stable per-seed IDs (never reused); the focus/highlight channel is a
  non-serialized display-node member (never scene XML), stripe phase animated
  via a custom event with zero MRML writes per tick.
- Pin = stripes only; **Restore placement view** is a separate, symmetric,
  guarded action (banner + "Return to my view", stale-snapshot refusal,
  placement guard); manual visibility edits always win.
- Hover-preview, divergence chip, slimmed rows with overflow menus, guarded
  destructive actions, pin persistence on the parameter node, idle-static
  stripes, on-slice pin text, ESC cancels an armed placement.
- Placement correctness fixes en route: labelmap-aimed pick (O(1), cached
  export), slice-only placement, immediate in-slice + 3D seed rendering,
  placement preview cursor, Show-3D preserving display parameters.

Docs: the data-first redesign + workflow critique design studies, in the
toctree.

Tests: extensive new bare + launched invariants (carve rule, carrier
grouping/IDs/contexts, gating truth table, presentation, pin/restore/preview
/persistence); launched suites are the behaviour gate in CI.

---

*Drafted with the assistance of Claude Code (model: claude-opus-4-8);
reviewed by the author.*


RafaelPalomar:
Appended the post-split eyeball round: icon-only Place/Pin toggles (stock Slicer icons, a11y names kept), and the fixes for the 11 launched failures this PR's first CI run surfaced — root causes: double-`setup()` in the widget-building test helpers leaking untracked scene observers (the destroyed-widget storm), missing observer detach-on-destroy guards in both module panels, plus per-test contract fixes. A few small VascularTerritories hardening hunks (destroy-guards, Place icon, test parents) ride along here rather than in #607 because they were discovered and fixed together; final tree is identical to the verified working branch.

RafaelPalomar:
Added a phantom-based **case-matrix suite** for the generation/compute algorithms (12 cases, analytic ground truth from the synthetic overlapping-segment phantom) after the maintainer flagged the generation algorithm as suspect. It proved Compute correct in every case and surfaced three real defects in Generate: (G1) generated segment names shifted one region on every run — the no-resection C++ branch labels the leftover 99, which imports last, while the naming loop assumed first; (G2) two seeds resolving to the same segment crashed (`GetNthSegment(i+1)` → None) and leaked a scratch node; (G3) Generate re-derived ownership from the flattened export (top-layer-wins) instead of each seed's binding + visibility snapshot, contradicting Compute's carve contract.

Generate's no-resection path is now rewritten onto the **same carved-region fold Compute uses** (`carved_masks_per_volume`), per the maintainer's contract decisions: one segment per volume (volume label + colour), cross-volume overlaps preserved as genuinely overlapping segments (matching Compute's independent per-volume rows), Unassigned = ticked export minus all claimed regions, ungrouped/empty-carve seeds skipped. The resection-refined path stays on the untouched C++ (ADR-0015); scratch-node leaks fixed on both paths. All 12 matrix cases now assert plain (no xfails), verified live.

RafaelPalomar:
Eyeball round: **temporary visualizations are now strictly module-scoped**, and fixing it uncovered the load-bearing cause.

The maintainer reported that seed/segment highlights survived switching to another module. Root cause (proven by live probing): `LiverVolumetry` and `VascularTerritories` are hidden modules whose panels the Liver shell composes via `widgetRepresentation()`, and the shell's `enter()`/`exit()` were `pass` — never forwarded. So **no** exit hygiene in either module had ever run in production: not the stripe retire, not the 25 Hz march-timer stop, not the module-active gate close. The vessel hover highlight was even still running surface picks and writing MRML on every mouse move under other modules.

Landed here: the Liver shell now relays `enter()`/`exit()` to the showing stage's panel (and across stage switches), targeting `page.self()` rather than `qSlicerAbstractModuleWidget::enter()` whose `IsEntered` assert would abort a Debug build; every interaction overlay in both modules (seed glyphs, slice handles, hover ring, placement preview, stripes + pinned-seed corner annotation, glow halo, adhering marker, vessel hover highlight) is hidden on exit and restored on enter, and each gate also vetoes grab arbitration so hidden handles stop answering clicks. Persistent results (centerline models, generated segmentations, the results table) and the user's own segment-visibility choices are deliberately untouched — asserted identical across an exit/enter round trip.

The drawing gate is split from the placement gate (opposite fail-safe directions) and its open value is a per-session nonce, so a scene saved with overlays up cannot resurrect them in a session where the module was never opened. Also: the Place toggle keeps one icon across both states (the markups `PlaceAdd` glyph means "add a point" in Slicer; the pushed state already signals armed). Verified live through real shell navigation and a fresh-instance scene load.

RafaelPalomar:
Eyeball round + CI fix:

- **Stripes march for the pin's whole life.** The march stopped after a fixed idle interval, which the maintainer read (correctly) as a rendering fault — a frozen texture is indistinguishable from a stuck view, so the motion is the "this overlay is live" signal. The countdown and its constant are gone; module `exit()` already stops the timer, so nothing ticks in the background.
- **Hover over the seed rows changes nothing.** Hovering a Pin button previewed that seed's stripes, so with a pin up the pinned region could not be told from whatever the cursor passed over — and a hover is an accident of moving the mouse. The Pin toggle and placement are now the only highlight drivers; the rewritten tests pin the inertness.
- **Place keeps one icon in both states** (the markups `PlaceAdd` glyph means "add a point" in Slicer; the pushed state already signals armed).
- **CI fix for the shell-lifecycle suite**: run from the repo root, `import Liver` resolves to the module DIRECTORY as a namespace package, which has no widget class, so all 11 tests failed on the missing attribute instead of skipping. Guarded like the sibling shell tests — 11 skipped from the root, 11 passed with the staged module path (where the contract is actually enforceable).

RafaelPalomar:
Restructured: the **Liver shell lifecycle relay** moved out of this PR into #609, which now sits between #607 and this one. It changes the shell rather than a stage module and fixes hygiene for every stage panel, so it deserves review on its own footing instead of being buried in a large volumetry PR. This PR's final tree is unchanged (verified byte-identical to the working branch after the rebase); the shell commits are simply in its ancestry via #609.

Stack: preview ← #607 (territory usability) ← #609 (shell lifecycle relay) ← #608 (volumetry workbench).

RafaelPalomar (commented):
LGTM

Issue #630, opened 2026-09-11
NURBS extension surface (ADR-0022) on the current parametric-surface base


RafaelPalomar:
Implement the NURBS extension surface (ADR-0022) against the current MRML architecture. This replaces PR #393 and issue #394, both closed 2026-09-11 — #393's foundations were superseded by #434 the day after it was last touched, so the work is cheaper to re-derive than to reconcile.

ADR-0022 is Accepted and was retargeted from v2.1 to v2.0 by the 2026-07-09 scope re-baseline. This issue gives that tracker checkbox something concrete behind it.

## Why the previous attempt cannot be rebased

- `vtkMRMLBezierSurfaceStorageNode`, which #393 extended, no longer exists — plan-rooted persistence goes through `vtkMRMLResectionPlanStorageNode`.
- #393 introduced `.lrp.json` **schema v3**; #434 trimmed the schema to **v2**, and preview now carries a negative assertion that `"schemaVersion": 3` is not emitted.
- `vtkMRMLAbstractParametricSurfaceNode` did not exist yet. It is now the required base for any parametric-surface node (ADR-0014 "Fourth layer"; ADR-0023 wrapper-vs-carrier).

## Scope

- `vtkMRMLNurbsSurfaceNode` deriving from `vtkMRMLAbstractParametricSurfaceNode`, alongside `vtkMRMLBezierSurfaceNode` rather than parallel to it.
- Persistence through the existing plan-rooted storage. **Decide first** whether NURBS needs a schema bump at all, or fits the current shape — #393 assumed a bump and that assumption is what aged worst.
- Registration per ADR-0013 §5; display via the shared `vtkMRMLParametricSurfaceDisplayNode` unless there is a reason not to (ADR-0014 says one display class serves many data subclasses — the Markups precedent).
- Invariant tests first per ADR-0039 §3/§4.

## Carried forward from #394 (found by review of the previous attempt — worth pinning as tests this time)

1. **Knot validation must not have an XML back door.** In #393, `ReadXMLAttributes` assigned `KnotsU`/`KnotsV` directly on a length match, bypassing `ValidateKnotsClampedMonotonic` — which guarded `SetKnotsU`/`SetKnotsV`/the JSON path but not XML. On failure the intended behaviour was a warning plus fallback to a clamped-uniform reset.
[...]

Issue #583, opened 2026-07-13
Extract a shared HandleGestureController from the three grab-grammar pipelines


RafaelPalomar:
## Problem

The handle-gesture interaction grammar is now implemented **three times**:

1. `LiverResectionsLib/ControlPolygonPipeline.py` — the Planning per-point drag (ADR-0033);
2. `LiverResectionsLib/LiverBezierSurfacePipeline.py` — the Init-handle drag (ADR-0032 / ADR-0035);
3. `LiverResectionsLib/SliceControlPolygonPipeline.py` — the slice-side control-point editing.

Each copy re-implements the same grammar:

- press within a display-space pick radius grabs one handle and returns the real squared distance for LayerDM arbitration; while grabbed, moves and the ending release are claimed unconditionally (`True, 0.0`);
- bare hover moves stay **declined** but raise the hover cue (halo + warm handle hue) as a side effect of the arbitration call;
- drag moves back-project at the grabbed handle's own depth (renderer World<->Display round-trip), never the focal plane;
- multi-field gesture writes batch under `StartModify`/`EndModify`;
- **the press itself must request a render** — the grab cue appears on the click, not at the first drag move (pinned by `test_press_repaints_before_any_drag_move` in two of the three copies; the init copy shipped without it and needed a fix on the #582 branch).

Visual constants (pick radius, halo colours/scale, grab/hover hues) are duplicated with "kept in sync by the pins" comments.

## Proposal

Extract a shared `HandleGestureController` owning the grammar above, parameterized by domain callbacks:

- `nearest_handle_in_display(renderer, eventData) -> (index, distance2)`
- `world_at_handle(renderer, eventData, index) -> (x, y, z)`
- `on_grab(index)` / `on_move(index, world)` / `on_release(index)` (the ADR-0035 domain-event raises live here)
- shared halo/hover choreography + the press-render rule built in, so the press-recolor class of bug becomes unrepresentable.

The three pipelines shrink to their callbacks; the interaction pins move to (or are complemented by) one controller-level suite.

## Open question: siting

The grammar is concept-agnostic — nothing in it is liver-specific. Should it live:

- in `LiverResectionsLib` (fast to iterate, stays repo-local), or
- upstream in the LayerDM SDK (SlicerLayerDisplayableManager), where every LayerDM extension writing grab-style pipelines would benefit?

A pragmatic path: land it in `LiverResectionsLib` first, propose upstream once the API survives a second consumer.

## Non-goals

[...]

Issue #505, opened 2026-07-06
T2.3 — ring-group control-point manipulation via the LayerDM Pipeline seam


RafaelPalomar:
## T2.3 — ring-group (grouped control-point) manipulation via the LayerDM Pipeline seam

The strategic payoff of moving the Bézier surface off Markups: markups have no point-group concept, so "move the corner / edge / interior ring of the 4×4 control grid as a unit" was impossible. This tracks that capability in the **post-ADR-0032** model. It is **new capability beyond v1 parity** — explicitly *not* part of #501 (the v1→v2 parity migration).

### Where it belongs (post-ADR-0032)
ADR-0014 §3 originally sited ring-group manipulation on a standalone `vtkLiverBezierWidget` (right-drag = ring-group; `TODO(T2.3 right-drag-ring-group)` + a `GroupDragging` state). **ADR-0032 superseded that widget** — interaction now routes through the LayerDM Pipeline seam. So ring-group re-homes to:

- **Mechanism** → a **right-drag branch on `LiverBezierSurfacePipeline.ProcessInteractionEvent`**, sibling to the slice-2 left-drag per-point edit (#501 slice 2). No standalone widget, no custom DisplayableManager (ADR-0013 §5).
- **GL-free kernel** → a Python method on the Pipeline, sibling to `_apply_world_point_to_nearest_control_point` — given a ring identity + a world delta, translate that ring's control points via `SetControlPoint`. Unit-testable GL-free (per slice 2's pattern); the routing verified on the interactive `:0` pass.
- **Ring taxonomy** (which flat indices form corners 4 / edges 8 / interior 4 for a `rows × cols` grid) → the **Algorithm library** `vtkSlicerLiverBezierControlPolygonGeometry` (pure VTK, ADR-0015 §1), a `BuildRingGroups(rows, cols)` sibling to the existing `BuildControlPolygonCells(rows, cols)`. Shared SSOT because **ADR-0018 says the ring taxonomy generalizes to NURBS** — both surfaces must consume one definition, not a Bézier-only Python helper.
- Right-click context menu (ring selection / mode switches) — the `TODO(T2.3 right-click-context-menu)` sibling — is the third interaction mode; also a Pipeline-seam concern (or a Python context menu per ADR-0004).

### ⚠️ Coordination risk (do this before the widget is deleted)
[...]

PR #636, opened 2026-09-11
ENH: Add BuildRingGroups to the control-polygon geometry SSOT (#505 slice 1)

RafaelPalomar:
> [!IMPORTANT]
> **CLOSED — not merged. Superseded by #639.**
>
> **Reason:** the API in this PR is not callable from Python, which is where both of its consumers live.
>
> `BuildRingGroups` returned `std::vector<std::vector<vtkIdType>>`. VTK's Python wrapper handles a flat
> `std::vector<T>` of scalars but **not a nested one**, so it skipped the method silently — no error, no
> warning, a green build. Verified against the generated wrapper in a real build:
>
> | Method | Return type | In `vtkSlicerLiverBezierControlPolygonGeometryPython.cxx` |
> |---|---|---|
> | `BuildControlPolygonCells` | `vtkSmartPointer<vtkCellArray>` | **yes** — registered in the method table |
> | `BuildRingGroups` | `std::vector<std::vector<vtkIdType>>` | **zero occurrences** |
>
> The C++ tests below all passed precisely because they are C++; they could not detect this. Both intended
> callers — `ControlPolygonPipeline` and `SliceControlPolygonPipeline` — are Python, so the method would have
> been unreachable, surfacing in slice 4 as "no such method".
>
> Fixing the signature alone (`vtkIdList` out-parameters wrap cleanly) was rejected on a second ground raised
> in review: ADR-0018 §1 admits only `{(3,3), (4,4)}`, each has exactly two rings, and the init re-fit only
> ever produces 4×4 — so the memberships are two fixed arrays. A wrapped C++ API plus an out-parameter loop
> to hand Python two constants is machinery without a payer.
>
> **#639** replaces this with module-level constants in `LiverResectionsLib`, where both pipelines import
> them: +159 lines of Python instead of +238 of C++ and generated wrapper surface, 10 tests passing **bare**,
> and packaging parity asserted so it is installed at runtime rather than merely present in the tree.
>
> Should a NURBS control net of arbitrary *m*×*n* ever need computed rings (#630), that is a C++ API designed
> against a real requirement — with a wrappable signature chosen deliberately and a Python consumer proving
> reachability.
>
> The branch is retained for reference; nothing here was merged.

---

Slice 1 of the control-polygon group-drag epic (#505). Pure geometry — no interaction, no rendering, no behaviour change. The gestures in later slices consume this.

## What and why

The two requested gestures both translate a **set** of control points by a single delta and recompute the surface:

- dragging the polygon **frame** moves the whole control polygon, its points, and the derived surface;
- right-dragging a **ring** moves that ring's points — all by the same direction and magnitude, not to the same place.

Both need ring **membership**, which nothing exposed. The Algorithm library carried `BuildControlPolygonCells` for the edge topology but no equivalent for the ring decomposition, so every caller would have rolled its own index arithmetic.

`BuildRingGroups(rows, cols)` goes beside its sibling per ADR-0018, which requires control-polygon geometry be shared with NURBS rather than reimplemented as a Bézier-only Python helper. Same row-major indexing (`i * cols + j`), same ADR-0018 §1 closed-shape rejection.

## The property that matters

**The rings partition the grid.** A ring translates rigidly, so an id appearing in two rings would be displaced twice, and an id in none would be left behind — either tears the surface. The test asserts the partition explicitly for both shapes rather than only checking ring sizes.

| Shape | Rings | Sizes |
|---|---|---|
| 4×4 | 2 | outer 12, inner 4 |
| 3×3 | 2 | outer 8, inner **1** |

Two details later slices depend on:

**The outer ring is a contiguous clockwise walk** from the top-left (`0,1,2,3,7,11,15,14,13,12,8,4`), so a ring highlight can consume the id order directly as a closed path instead of re-deriving the traversal. The test pins the exact sequence, not just the membership.

**A 3×3 grid's inner "ring" is the single centre point.** A one-element rigid-translation group is well defined, but it is not a ring in any geometric sense, and a caller presenting a ring affordance should expect it. The UI only ever produces 4×4 today (init always re-fits to it), but the shape is legal under ADR-0018 and the library must be correct for it. Flagging it here so it does not surface later as a crash in a gesture handler.

## Verification

- `vtkSlicerLiverBezierRingGroupsTest` — **passed**. Covers 4×4, 3×3, the exact outer-ring traversal order, the partition invariant, and rejection of `(2,2)`, `(5,5)`, `(3,4)`, `(4,3)`, `(0,0)`, `(1,1)` with the expected warning.
- `vtkSlicerLiverBezierControlPolygonGeometryTest1` (sibling) — **still passed**, so the shared header change did not disturb it.
- Built and run in the warm Qt6 tree; re-verified after clang-format reflowed two lines, so what is committed is what was tested. The borrowed tree was restored clean.

## Sequencing note — a correction

An earlier read of this epic had **#583** (folding `LiverBezierSurfacePipeline`'s grab grammar into the shared base) as a hard prerequisite, on the grounds that both touch `SurfacePointPlacementPipeline3D`'s dispatch.

That is wrong. `LiverBezierSurfacePipeline` has **zero** imports of `SlicerLiverInteractionLib`, and its grab grammar operates on the two auto-seeded **Init-handles** in `STATE_INIT` — a different data model from the control grid, and in `Planning` it declines outright because ADR-0033 re-sited per-point editing onto `ControlPolygonPipeline`. Widening the base cannot disturb it. #583 remains worth doing and will fold into the wider grammar, but it does not gate this work.

## Remaining slices

2. Group hover/grab state on `vtkMRMLControlPolygonDisplayNode` (it carries only single-int `HoveredControlPoint` / `GrabbedControlPoint` today; group state must live on the display node for the highlight to stay coherent across views).
3. Edge-segment picking — `CanProcessInteractionEvent` currently measures distance to handles only — plus widening the base's left-button-only dispatch.
4. Frame drag + frame highlight.
5. Ring right-drag + ring highlight.

Slices 4 and 5 will want a `:0` eyeball: LayerDM render-and-interaction work is exactly where the harnesses stayed green through five invisible defects during #620.

---
_Authored with assistance from Claude (Anthropic Claude Opus 5)._

### DIFF
--- a/LiverResections/Algorithm/Testing/Cxx/CMakeLists.txt
+++ b/LiverResections/Algorithm/Testing/Cxx/CMakeLists.txt
@@ -18,6 +18,7 @@ set(KIT_TEST_SRCS
   vtkLiverSpheroidRingExtractorEdgeCasesTest.cxx
   vtkLiverSpheroidRingExtractorConsistencyTest.cxx
   vtkSlicerLiverBezierControlPolygonGeometryTest1.cxx
+  vtkSlicerLiverBezierRingGroupsTest.cxx
   vtkLiverResectogramAspectRatioTest.cxx
   vtkLiverResectogramPixelMappingTest.cxx
   )
@@ -43,6 +44,7 @@ SIMPLE_TEST(vtkLiverContourParameterizerEdgeCasesTest)
 SIMPLE_TEST(vtkLiverPlaneRingExtractorEdgeCasesTest)
 SIMPLE_TEST(vtkLiverSpheroidRingExtractorEdgeCasesTest)
 SIMPLE_TEST(vtkSlicerLiverBezierControlPolygonGeometryTest1)
+SIMPLE_TEST(vtkSlicerLiverBezierRingGroupsTest)
 
 # Stack-4 ring-extraction wiring — Constraint 1 (shader/extractor
 # parameter consistency) + Constraint 2 (extraction-granularity bound).
--- a/LiverResections/Algorithm/Testing/Cxx/vtkSlicerLiverBezierRingGroupsTest.cxx
+++ b/LiverResections/Algorithm/Testing/Cxx/vtkSlicerLiverBezierRingGroupsTest.cxx
@@ -1,0 +1,172 @@
+/*==============================================================================
+
+ Distributed under the OSI-approved BSD 3-Clause License.
+
+  Copyright (c) 2026, Oslo University Hospital. All rights reserved.
+
+  Tests for
+  ``vtkSlicerLiverBezierControlPolygonGeometry::BuildRingGroups``
+  — the ring decomposition the control-polygon group-drag gestures
+  consume (frame translation and per-ring translation).
+
+  Per ADR-0008 §2 these are C++ low-level ctkTest-driver tests with no
+  Slicer scene, no Qt, no rendering, no on-screen interactor.  The
+  builder is a pure index helper; the test exercises the static method
+  directly.
+
+  The load-bearing property is that the rings PARTITION the control
+  grid: a ring translates rigidly, so an id appearing in two rings
+  would be displaced twice and an id appearing in none would be left
+  behind, tearing the surface.  Every case below asserts the partition
+  explicitly rather than only checking sizes.
+
+  Coverage:
+
+   - ``(4, 4)``: two rings — outer 12 ids, inner 4 — and the outer ring
+     emitted as a CONTIGUOUS closed-path walk, so a caller can draw the
+     ring highlight straight from the id order.
+   - ``(3, 3)``: two rings — outer 8 ids, and an inner group holding the
+     SINGLE centre id.  A one-element rigid-translation group is well
+     defined even though it is not a ring geometrically; this pins the
+     degenerate shape so a caller presenting a ring affordance cannot be
+     surprised by it.
+   - Invalid shapes: the ADR-0018 §1 closed-set check rejects them and
+     returns an empty ring list, mirroring the sibling
+     ``BuildControlPolygonCells``.
+
+==============================================================================*/
+
+// LiverResections Algorithm includes
+#include "vtkSlicerLiverBezierControlPolygonGeometry.h"
+
+// VTK includes
+#include <vtkTestingOutputWindow.h>
+#include <vtkType.h>
+
+// STD includes
+#include <algorithm>
+#include <cstdio>
+#include <cstdlib>
+#include <iostream>
+#include <vector>
+
+namespace
+{
+
+#define LIVER_CHECK_INT(actual, expected)                                                                    \
+  do                                                                                                         \
+  {                                                                                                          \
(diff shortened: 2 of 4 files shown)


RafaelPalomar:
Closing: the signature is not Python-wrappable, and once corrected the C++ API no longer earns its place.

## The wrapping defect

`BuildRingGroups` returned `std::vector<std::vector<vtkIdType>>`. VTK's Python wrapper handles a flat `std::vector<T>` of scalars but **not a nested one**, so it skipped the method silently — no error, no warning, a green build.

Verified against the generated wrapper in a real build:

| Method | Return type | In `vtkSlicerLiverBezierControlPolygonGeometryPython.cxx` |
|---|---|---|
| `BuildControlPolygonCells` | `vtkSmartPointer<vtkCellArray>` | **yes** — 6 references, registered in the method table |
| `BuildRingGroups` | `std::vector<std::vector<vtkIdType>>` | **zero occurrences** |

Both consumers are Python (`ControlPolygonPipeline`, `SliceControlPolygonPipeline`), so neither could ever have called it. The C++ tests passed precisely because they are C++ — they could not detect this.

Same failure mode as the `void*` distance-map upload in #620: a signature that compiles, links, tests green, and is invisible from Python. Worth noting as a recurring trap rather than a one-off.

## Why not just fix the signature

`vtkIdList` out-parameters would wrap cleanly, and would be the idiomatic VTK choice. But ADR-0018 §1 admits only `{(3,3), (4,4)}`, each has exactly two rings, and the init re-fit only ever produces 4×4 — so the memberships are two fixed arrays. Adding a wrapped C++ API plus an out-parameter loop to hand Python two constants is machinery without a payer.

Replaced by module-level constants in `LiverResectionsLib`, where both pipelines import them.

If NURBS (#630) later needs computed rings for an arbitrary *m*×*n* control net, that is a C++ API designed against a real requirement — with a wrappable signature chosen deliberately, and a Python consumer to prove it.

Superseded by the follow-up PR; branch retained for reference.

---
_Triaged with assistance from Claude (Anthropic Claude Opus 5)._

RafaelPalomar:
Superseded by **#639**, now open.

My earlier comment on this thread said "superseded by the follow-up PR" before that PR existed — this names it, so the trail is followable from here.

What changed in the replacement, beyond the language:

- **It is reachable.** 10 tests pass **bare**, with no build, which is the direct evidence the original lacked — the C++ version needed a compile and still could not be called from where it was used.
- **Packaging parity is asserted.** The new module is registered in `LiverResectionsLib_PYTHON_SCRIPTS` and `test_liverresectionslib_packaging_parity` covers it. Skipping that step would have reproduced this PR's defect in a different costume: present in the source tree, absent at runtime.
- **The invariants carried over unchanged** — the partition assertion, the exact clockwise outer-ring order the highlight draws from, the 3×3 single-centre-point degenerate case, and empty-on-out-of-spec-shape.

The general peel-the-lattice algorithm this PR contained is *not* carried over; #639 uses constants, per the review point that ADR-0018 §1 leaves only two legal shapes with two rings each.

PR #649, opened 2026-09-12
ENH: Offer Poisson reconstruction alongside marching cubes (ADR-0040 phase 4a)

RafaelPalomar:
**ADR-0040 phase 4a.** The consumer seam: a segmentation segment becomes a mesh **by a method the caller names**. Stacked on **#648**; also merges **#645** (the extractor), because the consumer composes *both* primitives and they sit on parallel branches. Relates to **#310**.

**Nothing changes for anyone who does not ask.** `ensure_target_model` gains a `method` parameter whose absence means marching cubes, so the liver target mesh is byte-for-byte what it was.

## The choice is the deliverable

ADR-0040 §Conformance requires PSR be offered *alongside* marching cubes, never as a silent replacement — and the reason isn't caution, it's that the two make a trade the surgeon owns:

- **Marching cubes** is exact about which voxels are in, and inherits the labelmap's voxel staircase.
- **PSR** approximates rather than interpolates, and will not reproduce a one-voxel spike.

At CT resolution that staircase is a real feature of the data, not noise to be assumed away. Neither is better in general, so this module must not decide.

**An unknown method returns `None` rather than falling back.** A caller who asked for PSR and silently got marching cubes would be comparing the two and reading the same surface twice — the failure mode that makes an A/B evaluation worthless while looking like it worked.

## Two details that the composition has to get right

**The label value is threaded, not just the image.** Stage 2's segments share one binary-labelmap layer, so the image a segment hands back holds its neighbours' voxels; only the segment's own value distinguishes it. That's why the extractor tests voxels for *equality*. A test with two concentric labelled shells pins that the composition preserves it end to end — reconstructing label 3 must not reach into the label-7 shell wrapped around it.

**The image-to-world matrix is passed whole.** A segment's labelmap is oriented — spacing, direction and origin together — so handing the extractor the matrix rather than a spacing triple is what stops an oblique acquisition being flattened into an axis-aligned assumption.

`_liver_closed_surface` stays as a thin alias to the moved implementation: once a second method exists, marching cubes has no claim to being the unqualified meaning of "the surface".

## Tests run in both rows, and the split is deliberate

The **dispatch** half — both methods on the menu, marching cubes still first and still the default, an unrecognised method refused — is the ADR conformance invariant, and it **runs in the bare row** rather than skipping. The fixture falls back to loading `SegmentSurface` from its source path, which is safe *only* because the module imports `vtk` and nothing else at module scope and resolves the wrapped classes lazily. If that stops being true, the fixture fails loudly instead of quietly skipping.

That mattered: my first attempt registered a bare row that skipped all 9 tests, because `LiverResectionsLib` isn't importable outside a launched Slicer. A green row that only ever skips is worth nothing, and this project has paid for that lesson repeatedly (#449, #454, #459, #460).

The default is asserted **through the real dispatcher with a recording stub**, not by reading the signature's default — so a later refactor that moves the default elsewhere still trips it.

```
bare      4 passed,  5 skipped
launched  9 passed
full launched LiverResections root   216 passed, 15 skipped (all pre-existing)
Algorithm C++ suite                  15/15
bare pytest_scaffold                 green
```

## Next

Phase 4b: the UI that lets the surgeon make the choice, and the `:0` eyeball comparing the two on real anatomy — which is where the trade above stops being an argument and becomes something you can look at.

---
_Authored with assistance from Claude (Anthropic Claude Opus 5)._

### DIFF
--- a/LiverResections/Algorithm/CMakeLists.txt
+++ b/LiverResections/Algorithm/CMakeLists.txt
@@ -62,6 +62,8 @@ set(${KIT}_SRCS
   vtkLiverResectogramAspectRatio.cxx
   vtkLiverResectogramPixelMapping.h
   vtkLiverResectogramPixelMapping.cxx
+  vtkLiverOrientedPointCloudExtractor.h
+  vtkLiverOrientedPointCloudExtractor.cxx
   )
 
 #
--- a/LiverResections/Algorithm/Testing/Cxx/CMakeLists.txt
+++ b/LiverResections/Algorithm/Testing/Cxx/CMakeLists.txt
@@ -18,6 +18,7 @@ set(KIT_TEST_SRCS
   vtkLiverSpheroidRingExtractorEdgeCasesTest.cxx
   vtkLiverSpheroidRingExtractorConsistencyTest.cxx
   vtkSlicerLiverBezierControlPolygonGeometryTest1.cxx
+  vtkLiverOrientedPointCloudExtractorTest.cxx
   vtkLiverResectogramAspectRatioTest.cxx
   vtkLiverResectogramPixelMappingTest.cxx
   )
@@ -43,6 +44,7 @@ SIMPLE_TEST(vtkLiverContourParameterizerEdgeCasesTest)
 SIMPLE_TEST(vtkLiverPlaneRingExtractorEdgeCasesTest)
 SIMPLE_TEST(vtkLiverSpheroidRingExtractorEdgeCasesTest)
 SIMPLE_TEST(vtkSlicerLiverBezierControlPolygonGeometryTest1)
+SIMPLE_TEST(vtkLiverOrientedPointCloudExtractorTest)
 
 # Stack-4 ring-extraction wiring — Constraint 1 (shader/extractor
 # parameter consistency) + Constraint 2 (extraction-granularity bound).
--- a/LiverResections/Algorithm/Testing/Cxx/vtkLiverOrientedPointCloudExtractorTest.cxx
+++ b/LiverResections/Algorithm/Testing/Cxx/vtkLiverOrientedPointCloudExtractorTest.cxx
@@ -1,0 +1,286 @@
+/*==============================================================================
+
+ Distributed under the OSI-approved BSD 3-Clause License.
+
+  Copyright (c) 2026, Oslo University Hospital. All rights reserved.
+
+  Tests for ``vtkLiverOrientedPointCloudExtractor`` — the oriented point
+  cloud PSR consumes (ADR-0040 §Decision 2).
+
+  The load-bearing property is the NORMAL DIRECTION.  PSR reconstructs
+  from orientation; a cloud whose normals point inward reconstructs the
+  complement of the object, and a cloud with the right positions but
+  wrong orientations fails in a way position-only assertions cannot see.
+  So the sphere case checks every normal against the analytic outward
+  direction, not merely that normals exist.
+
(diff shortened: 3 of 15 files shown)


RafaelPalomar:
---

## Update — the `:0` eyeball ran, and found a bug (2026-09-13)

Reconstructing a real patient's liver put the Poisson surface **53.179 mm mean** from the marching-cubes surface *of the same voxels*.

**Cause:** the extractor dropped the labelmap's **extent origin**. A segment's binary labelmap is cropped to its own bounding box, so its extent never starts at zero, and Slicer's image-to-world matrix maps *absolute* voxel indices. The cloud came out intact, correctly shaped and correctly oriented — just translated.

**Five C++ tests passed throughout.** Normals-vs-analytic-outward, deterministic subsampling, geometry-from-the-matrix, degenerate inputs, two-label no-bleed. Every one built its fixture with `SetDimensions`, which yields a zero-based extent, where a dropped offset is indistinguishable from a correct one. The assertions were strong; the inputs were the wrong *shape*.

Fixed on #645 (`BUG: Honour the labelmap's extent origin when emitting points`), with `TestExtentOriginIsHonoured` verified **red before green** — without the fix, "axis 0 off by 100".

After the fix, on the real liver:

| | points | polys | time |
| --- | --- | --- | --- |
| marching cubes | 145,150 | 290,296 | 0.10 s |
| poisson depth 7 | 50,598 | 101,192 | 14.3 s |

```
mc  -> psr : mean 0.068  worst 0.982 mm
psr -> mc  : mean 0.078  worst 1.020 mm
0 degenerate normals on either surface
```

PSR carries the same surface in **a third of the triangles**, agreeing to 0.068 mm mean — so the difference is concentrated in the staircase, which is exactly what a surgeon would be choosing between.

One rendering note for anyone repeating this: the first comparison used flat shading to expose the staircase, and **inverted the result** — flat shading discards per-vertex normals, so the surface with fewer, larger triangles facets more, and PSR looked rougher than marching cubes. The eyeball is only evidence if the shading is matched. The solver's normals were independently verified: all unit length, none inverted, `dot = 1.0000` against normals derived from the output triangles.

## Phase 4b, also in this PR

Maintainer decisions: PSR feeds the **target mesh** as well as the display, and the choice lives in **Stage 2**.

`apply_surface_method` writes the chosen mesh **into the segment's closed-surface representation** — the one seam every downstream consumer already reads. The alternative was threading a `method` argument to the 3D display, to `TargetModel`, and through it to ring extraction. Installing at the representation means they inherit the choice unchanged, and makes it *impossible* for the rendered anatomy and the geometry the plan is computed against to disagree — a discrepancy a surgeon would have no way to see, with nothing in the UI wrong.

Probed before building, not assumed:

| | result |
| --- | --- |
| read back after injection | **held** |
| after `CreateClosedSurfaceRepresentation()` | **held** — Slicer won't regenerate an existing representation |
| after a labelmap edit | **regenerated** |

That last row is the right way round: a reconstruction of voxels the user has since changed would be stale, and looking correct is what makes stale dangerous. So the tag records what was *last applied*, not a guarantee of what is there, and says so.

The tag is on the **segment**, not the node — a case may want a smooth liver and a staircase tumour. `applied_surface_method` returns `None` for "nobody chose", distinct from "marching cubes was chosen", so the Stage-2 UI can show an honest initial value rather than asserting a default the user never picked.

The scene fixture builds its labelmap at a **non-zero extent**, for the reason above.

```
launched  13 passed
bare       4 passed, 5 skipped
Algorithm C++ suite  15/15
```

Still to come in 4c: the Stage-2 control itself, and an eyeball of the choice driving a real resection.

